PR #9, opened 2024-10-09
Prepare for Python bindings

aliddell:
- Change the CMake option `NOTEST` (a negative) to `BUILD_TESTS` (a positive), defaults to `ON`
  - Conditionally include `tests` directory in the base CMakeLists.txt file based on this option
- Add a bunch of Python-related stuff to .gitignore
- Set `POSITION_INDEPENDENT_CODE` to `ON` in the streaming library
- Fix under-indent in acquire.zarr.h

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,12 +93,12 @@ jobs:
 
       - name: Build for x64
         run: |
-          cmake --preset=default -DVCPKG_TARGET_TRIPLET=x64-osx -DVCPKG_INSTALLED_DIR=${{github.workspace}}/vcpkg-x64 -B ${{github.workspace}}/build-x64 -DCMAKE_BUILD_TYPE=${{matrix.build_type}} -DCMAKE_OSX_ARCHITECTURES="x86_64" -DNOTEST=1
+          cmake --preset=default -DVCPKG_TARGET_TRIPLET=x64-osx -DVCPKG_INSTALLED_DIR=${{github.workspace}}/vcpkg-x64 -B ${{github.workspace}}/build-x64 -DCMAKE_BUILD_TYPE=${{matrix.build_type}} -DCMAKE_OSX_ARCHITECTURES="x86_64" -DBUILD_TESTS=OFF
           cmake --build ${{github.workspace}}/build-x64 --config ${{matrix.build_type}}
 
       - name: Build for arm64
         run: |
-          cmake --preset=default -DVCPKG_TARGET_TRIPLET=arm64-osx -DVCPKG_INSTALLED_DIR=${{github.workspace}}/vcpkg-arm64 -B ${{github.workspace}}/build-arm64 -DCMAKE_BUILD_TYPE=${{matrix.build_type}} -DCMAKE_OSX_ARCHITECTURES="arm64" -DNOTEST=1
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.23)
+cmake_minimum_required(VERSION 3.30)
```
jeskesen, line 1: Is there a feature you used that required this version upgrade? I'd like it if we could keep this version as low as possible. 

From my own selfish (ROS2) perspective, 3.23 already breaks the build for ubuntu 22.04 (which has CMake v3.22) , and anything higher than 3.28 will break it for 24.04.    One can always go and install more up-to-date CMake versions using kitware's PPA, but it's just another barrier to entry I'd like to avoid, if possible.

CMakeLists.txt
```diff
@@ -19,9 +19,13 @@ set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
-option(NOTEST "Disable all tests" OFF)
+option(BUILD_TESTS "Build tests" ON)
```
jeskesen, line 22: Do you think, rather than having our own flag, we can use the built-in method for enabling the build of tests?  https://cliutils.gitlab.io/modern-cmake/chapters/testing.html

jeskesen (changes requested):
Lets at least talk about the comments, before proceeding.

.github/workflows/build.yml
```diff
@@ -93,12 +93,12 @@ jobs:
...
       - name: Build for x64
         run: |
-          cmake --preset=default -DVCPKG_TARGET_TRIPLET=x64-osx -DVCPKG_INSTALLED_DIR=${{github.workspace}}/vcpkg-x64 -B ${{github.workspace}}/build-x64 -DCMAKE_BUILD_TYPE=${{matrix.build_type}} -DCMAKE_OSX_ARCHITECTURES="x86_64" -DNOTEST=1
+          cmake --preset=default -DVCPKG_TARGET_TRIPLET=x64-osx -DVCPKG_INSTALLED_DIR=${{github.workspace}}/vcpkg-x64 -B ${{github.workspace}}/build-x64 -DCMAKE_BUILD_TYPE=${{matrix.build_type}} -DCMAKE_OSX_ARCHITECTURES="x86_64" -DBUILD_TESTS=OFF
           cmake --build ${{github.workspace}}/build-x64 --config ${{matrix.build_type}}
 
       - name: Build for arm64
         run: |
-          cmake --preset=default -DVCPKG_TARGET_TRIPLET=arm64-osx -DVCPKG_INSTALLED_DIR=${{github.workspace}}/vcpkg-arm64 -B ${{github.workspace}}/build-arm64 -DCMAKE_BUILD_TYPE=${{matrix.build_type}} -DCMAKE_OSX_ARCHITECTURES="arm64" -DNOTEST=1
+          cmake --preset=default -DVCPKG_TARGET_TRIPLET=arm64-osx -DVCPKG_INSTALLED_DIR=${{github.workspace}}/vcpkg-arm64 -B ${{github.workspace}}/build-arm64 -DCMAKE_BUILD_TYPE=${{matrix.build_type}} -DCMAKE_OSX_ARCHITECTURES="arm64" -DBUILD_TESTS=OFF
```
jeskesen: It looks like there are several references to  BUILD_TESTS that were missed?    Maybe these don't even need to get set anymore?

aliddell: Yeah I remembered these after I requested another review.

jeskesen (approved):
Thanks for making those updates!

PR #10, opened 2024-10-09
Implement Python bindings

aliddell:
- Implement Python bindings
  - Add a CMake option (off by default) to build Python bindings
  - Add setup.py, pyproject.toml, and MANIFEST.in with which you can install acquire_zarr in your environment
  - Add a job to build Python wheels

Tests coming later. Stay tuned...


jeskesen:
LGTM, but I'd like to get someone else's opinion on the Python packaging, because I don't know that part.

python/acquire-zarr-py.cpp
```diff
@@ -0,0 +1,443 @@
...
+           :toctree: _generate
+           append
+    )pbdoc";
+
+    py::enum_<ZarrVersion>(m, "Version")
+      .value("V2", ZarrVersion_2)
+      .value("V3", ZarrVersion_3)
+      .export_values();
+
+    py::enum_<ZarrDataType>(m, "DType")
```
jeskesen, line 297: A topic for discussion: do we think it might be worth building ways to implicitly (or explicitly) convert between numpy's DType & ZarrDataTypes?

jeskesen (approved):
So, from my perspective, it wasn't readily understandable how to build the new python bindings.   I did it the "CMake way", with a configure/build/install step, and they weren't installed in the "install" directory (as it appears there's no install step added in the 

Over Slack, @aliddell showed me how to build it the "python" way: ` python -m build`, and that worked fine.  So, I'll approve this, but can we open a new ticket to update build instructions?

python/acquire-zarr-py.cpp
aliddell: Yeah, I think that might be worthwhile to do.

PR #11, opened 2024-10-10
Add Python tests (1)

aliddell:
- Adds some basic tests for the settings classes
  - Adds Python tests to CI
- Fixes binding of `PyZarrStreamSettings::dimensions` to allow list-like behavior in Python
- Adds default values for previously uninitialized class members
- Adds `kwargs` constructors and `__repr__` to bound classes


jeskesen (approved):
LGTM.

PR #17, opened 2024-11-05
Set log level from Python

aliddell:
- Suppresses log messages below on set log level
- Adds a new log level, `LogLevel_None`, for no logging
- Exposes `LogLevel` enum, `get_log_level`, and `set_log_level` in Python


jeskesen (approved):
LGTM

PR #20, opened 2024-11-27
Vectorized file writes

aliddell:
Adds a class which takes a `std::vector<std::span<std::byte>>>&` and writes the contents of each span in a vectorized fashion to the filesystem. Includes some system-dependent code because of course it does.

Also adds benchmarking code so that we can test streaming performance before and after this change.


jeskesen (changes requested):
It looks like there's a lot of duplicated code between ChunkIndexBuffer & ShardIndexBuffer.  Could they be combined?

aliddell:
> It looks like there's a lot of duplicated code between ChunkIndexBuffer & ShardIndexBuffer. Could they be combined?

Actually they don't belong there. Those are artifacts from something I was trying but discarded.

jeskesen (approved):
LGTM

PR #21, opened 2024-12-04
Add some examples.

aliddell:
C and Python examples, tested working.


examples/python/zarrv2-raw-filesystem.py
```diff
@@ -0,0 +1,55 @@
...
+
+    settings.store_path = "output_v2.zarr"
+    settings.version = ZarrVersion.V2
+    settings.data_type = DataType.INT32
+
+    # Create stream
+    stream = ZarrStream(settings)
+
+    # Create and write sample data
+    for i in range(10):
```
jeskesen, line 45: This is just a (nit-picky) suggestion, that we should discuss:
For examples, it might make it more intuitive to show the calls closer to how they actually would be used.  Here, you generate the data in the same loop as you write it.  It might it more intuitive to generate all the data up front and then have only the append (which indexes into an array) inside the loop.   Another possibility would be to put the image generation call inside a get_image_data function call (which gets called within the loop).

jeskesen: This comment applies to all the python examples (won't repeat for each)

aliddell:
Also closes #25.

jeskesen (approved):
Thanks for addressing those comments.  LGTM.

PR #23, opened 2024-12-05
Vectorized chunk writes

aliddell:
- Add a few virtual methods to `ArrayWriter`
  - `data_root_()` constructs the path to the location where chunk or shard files are currently going to be written
  - `metadata_path_()` constructs the path to the array metadata file
  - Remove `version_()`
  - Replace `compress_buffers_()` with `compress_chunk_buffer_()`, which compresses one buffer at a time
  - Make `make_data_sinks_()` and `close_sinks_()` pure virtual methods
- Rename `SinkCreator` methods to be explicit about which type of sink (S3 or filesystem).
- Zarr V2: Enqueue each chunk to be written immediately after compression, rather than waiting until all chunks are compressed before writing them
- Zarr V3: Write chunks to shards using `FileWriteGather` (Windows) or `pwritev` (POSIX).


src/streaming/array.writer.cpp
```diff
@@ -400,15 +323,12 @@ zarr::ArrayWriter::flush_()
...
-    CHECK(flush_impl_());
+    compress_and_flush_();
 
     const auto should_rollover = should_rollover_();
-    if (should_rollover) {
-        rollover_();
-    }
 
     if (should_rollover || is_finalizing_) {
+        rollover_();
```
jeskesen, line 331: Compared to the old functionality, rollover_() would now be called if should_rollover==false but is_finalizing==true, whereas it wouldn't have before.  I just want to make sure that is intended.

aliddell: Correct. I previously didn't want to call `rollover_()` when finalizing because it would uselessly increment  a counter, but the alternative was to add another check on `is_finalizing_` to call `close_files_()` -- but only when we had a ragged append dimension. This is simpler, and the increment doesn't matter.

jeskesen (approved):

src/streaming/array.writer.cpp
```diff
@@ -138,84 +137,21 @@ zarr::ArrayWriter::is_s3_array_() const
...
-                            ".array.json";
-            break;
-        default:
-            LOG_ERROR("Unsupported Zarr version");
-            return false;
-    }
+    const auto metadata_path = metadata_path_();
 
     if (is_s3_array_()) {
         SinkCreator creator(thread_pool_, s3_connection_pool_);
```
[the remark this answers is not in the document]
aliddell: I'll just remove the namespace identifier from line 154.

PR #31, opened 2024-12-17
Convert acquire-logger from a static library to an object library.

### LINKED ISSUE #28: logger library should not be required by clients
The logger library (src/logger) gets built as a static library `acquire-logger`. 

When someone wants to depend on the `acquire-zarr` static library, they need to also link `acquire-logger`. But that library is only providing definitions for private symbols in `acquire-zarr` - it's not required for any part of that library's public interface.

Here's a proposed approach:

Ideally, `acquire-zarr` would be built in such a way that all the `acquire-logger` symbol definitions are all included. We could do this by listing the source files in src/logger directly or maybe by exposing on object library target. I believe cmake effectively adds the object files when compiling the lib which would resolve the issue.

aliddell:
Closes #28.

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -11,7 +11,6 @@ target_include_directories(${tgt} PRIVATE
         ${PROJECT_SOURCE_DIR}/src/logger
 )
 target_link_libraries(${tgt} PRIVATE
-        acquire-logger
         acquire-zarr
         nlohmann_json::nlohmann_json
         miniocpp::miniocpp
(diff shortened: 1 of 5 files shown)


tlambert03:
awesome!!  This works for me.  

For those building with meson, with this PR i can now easily build against acquire-zarr by just including this wrap in `subprojects/acquire-zarr.wrap`:

```ini
[wrap-git]
url = https://github.com/acquire-project/acquire-zarr
revision = 28-private-logger-library  # or later something else
depth = 1
method = cmake
```

and then adding `aquire_zarr_dep = dependency('acquire-zarr')` to my meson.build

thanks @aliddell

jeskesen (approved):
LGTM.   would it make sense to add @lambert03's commen wrapper to the README?

nclack (approved):
LGTM!

PR #43, opened 2025-01-15
Compress and flush

aliddell:
Previous workflow was to use n threads to compress n buffers, then once this was done, use n threads again to write n buffers out to disk (or to S3). New workflow is to use compress and write each buffer sequentially in the same thread worker, so that, in principle, a single chunk can be compressed and written out before another chunk begins compressing. This should reduce thread synchronization overhead.


jeskesen (approved):
Discussed over zoom

PR #45, opened 2025-01-17
Remove ArrayWriter::version_()

aliddell:
- Removes `ArrayWriter::version_()`
- Creates `ArrayWriter::data_root_()`, `ArrayWriter::metadata_path_()`, and `ArrayWriter::parts_along_dimension_()`

Gets rid of some ugly code that depended on the output of `version_()` in switch statements. Instead, each subclass must implement the new methods.


jeskesen (approved):
LGTM.

PR #47, opened 2025-01-22
Add make_dirs and construct_data_paths to sink.cpp

aliddell:
Moving away from the `SinkCreator` class. Depends on #45.

### DIFF
--- a/src/streaming/sink.cpp
+++ b/src/streaming/sink.cpp
@@ -1,5 +1,15 @@
 #include "sink.hh"
+#include "file.sink.hh"
+#include "s3.sink.hh"
 #include "macros.hh"
+
+#include <algorithm>
+#include <filesystem>
+#include <latch>
+#include <stdexcept>
+#include <unordered_set>
+
+namespace fs = std::filesystem;
 
 bool
 zarr::finalize_sink(std::unique_ptr<zarr::Sink>&& sink)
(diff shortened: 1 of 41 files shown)


src/streaming/sink.cpp
```diff
@@ -15,4 +25,111 @@ zarr::finalize_sink(std::unique_ptr<zarr::Sink>&& sink)
...
+
+            std::error_code ec;
+            if (!fs::create_directories(path, ec)) {
+                err =
+                  "Failed to create directory '" + path + "': " + ec.message();
+                success = false;
+            }
+
+            latch.count_down();
+            all_successful.fetch_and(static_cast<char>(success));
```
nclack, line 121: fwiw, you can use release memory ordering here and acquire on the final read.

The default (cst) should work fine and I would be surprised if it mattered for performance.

src/streaming/sink.cpp
```diff
@@ -15,4 +25,111 @@ zarr::finalize_sink(std::unique_ptr<zarr::Sink>&& sink)
 
     sink.reset();
     return true;
-}
\ No newline at end of file
+}
+
+std::vector<std::string>
+zarr::construct_data_paths(std::string_view base_path,
```
nclack, line 31: Is this function used anywhere outside the tests? My guess is the actual use comes in a follow-up pr :)

In that case, what's the difference between construct_data_paths and make_dirs? It's hard to tell from the function names. 

I can't tell how construct_data_paths will be used, but is it possible needing to store all the paths in a vector? e.g. Can you pass a path handler as an argument and then this function calls the handler for every constructed path.

nclack (approved):
I had a couple questions. One thing that might help me for these prs is to point out where you want review - I suspect I'm commenting/asking about things that have already been discussed in the past. Nothing major so approving.

src/streaming/sink.cpp
aliddell: I see that I missed documenting the return value for `construct_data_paths`. This constructs the paths (which can be used to create directories or S3 object keys), but doesn't actually create the directories.

src/streaming/sink.cpp
```diff
@@ -15,4 +25,111 @@ zarr::finalize_sink(std::unique_ptr<zarr::Sink>&& sink)
...
+bool
+zarr::make_dirs(const std::vector<std::string>& dir_paths,
+                std::shared_ptr<ThreadPool> thread_pool)
+{
+    if (dir_paths.empty()) {
+        return true;
+    }
+    EXPECT(thread_pool, "Thread pool not provided.");
+
+    std::atomic<char> all_successful = 1;
```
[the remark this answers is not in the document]
aliddell: `std::atomic<bool>` was my first choice here, but it doesn't support `fetch_and`, if you can believe it.

PR #49, opened 2025-01-22
Deprecate `SinkCreator`

aliddell:
Paving the way for writing to non-`Sink` classes. Depends on #47.


src/streaming/sink.cpp
```diff
@@ -11,6 +12,224 @@
...
+    return (bool)all_successful;
+}
+
+bool
+make_s3_sinks(std::string_view bucket_name,
+              const std::vector<std::string>& object_keys,
+              std::shared_ptr<zarr::S3ConnectionPool> connection_pool,
+              std::vector<std::unique_ptr<zarr::Sink>>& sinks)
+{
+    if (object_keys.empty()) {
```
jeskesen, line 178: Is this check necessary?  The for loop will just be skipped at the bottom.  Is it so the other error checks get skipped if you passed in an empty vector?

src/streaming/sink.cpp
```diff
@@ -11,6 +12,224 @@
...
+
+bool
+make_s3_sinks(
+  std::string_view bucket_name,
+  std::string_view base_path,
+  const std::vector<std::string>& object_keys,
+  std::shared_ptr<zarr::S3ConnectionPool> connection_pool,
+  std::unordered_map<std::string, std::unique_ptr<zarr::Sink>>& sinks)
+{
+    if (object_keys.empty()) {
```
jeskesen, line 209: same question as above.

tests/unit-tests/CMakeLists.txt
```diff
@@ -17,8 +17,8 @@ set(tests
         file-sink-write
         s3-sink-write
         s3-sink-write-multipart
-        sink-creator-make-metadata-sinks
```
jeskesen, line 20: should sink-creator-* files get removed?

tests/unit-tests/CMakeLists.txt
```diff
@@ -17,8 +17,8 @@ set(tests
...
-        sink-creator-make-data-sinks
+        make-data-sinks
+        make-metadata-sinks
```
jeskesen, line 21: There are 2 new common-* that have not been added to CMakeLists.txt file.  Intentional?

jeskesen (approved):
Some non-blocking comments...

tests/unit-tests/CMakeLists.txt
aliddell: Nope, good catch.

PR #51, opened 2025-01-23
Test partial appends. Fix buffer overrun on partial append.

aliddell:
This PR adds a test for partial appends -- and a good thing, because there was a bug in the logic which could lead to a buffer overrun.

### DIFF
--- a/tests/unit-tests/partial-append.cpp
+++ b/tests/unit-tests/partial-append.cpp
@@ -1,0 +1,230 @@
+#include "acquire.zarr.h"
+#include "zarr.stream.hh"
+#include "unit.test.macros.hh"
+
+#include <filesystem>
+#include <fstream>
+
+namespace fs = std::filesystem;
+
+void
+configure_stream_dimensions(ZarrStreamSettings* settings)
+{
+    CHECK(ZarrStatusCode_Success ==
+          ZarrStreamSettings_create_dimension_array(settings, 3));
+    ZarrDimensionProperties* dim = settings->dimensions;
+
+    *dim = ZarrDimensionProperties{
+        .name = "t",
+        .type = ZarrDimensionType_Time,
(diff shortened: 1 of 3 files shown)


tests/unit-tests/partial-append.cpp
```diff
@@ -0,0 +1,230 @@
...
+        .array_size_px = 64,
+        .chunk_size_px = 64,
+    };
+}
+
+void
+verify_file_data(const ZarrStreamSettings& settings)
+{
+    std::vector<uint8_t> buffer;
+    const size_t row_size = 64, num_rows = 48;
```
jeskesen, line 45: is there a benefit to initializing this separately from the dimensions stored in ZarrStreamSettings?

tests/unit-tests/partial-append.cpp
```diff
@@ -0,0 +1,230 @@
...
+                                       data.size() - bytes_to_write);
+        EXPECT_EQ(int, data.size() - bytes_to_write, bytes_written);
+
+        retval = 0;
+    } catch (const std::exception& exception) {
+        LOG_ERROR(exception.what());
+    }
+
+    // cleanup
+    ZarrStream_destroy(stream);
```
jeskesen, line 215: why not combine the try catch blocks and move this call to the end?

jeskesen (approved):
A couple of comments in the test code... certainly not blocking

tests/unit-tests/partial-append.cpp
aliddell: Nope, just an oversight.

tests/unit-tests/partial-append.cpp
aliddell: I was debugging something and forgot to put this back. Good catch.

PR #52, opened 2025-01-24
Write a vector of buffers from a `Sink` (deprecate `VectorizedFileWriter`)

aliddell:
Depends on #49.


nclack:
@aliddell this should be closed in favor of #60?

PR #54, opened 2025-02-06
Fix wheel to include stubs

tlambert03:
currently, the wheel doesn't include the pyi file, so there's no static typing support when you pip install acquire-zarr.  This fixes it, but it does so by making it a package (rather than a single-module distribution).  I struggled with setuptools for a bit to try to add the .pyi as a top level data file alongside the compiled single module, but failed...


aliddell (approved):
Thanks!

PR #58, opened 2025-02-10
Add `region` field to S3 settings

aliddell:


src/streaming/s3.connection.cpp
```diff
@@ -21,6 +21,22 @@ zarr::S3Connection::S3Connection(const std::string& endpoint,
     CHECK(client_);
 }
 
+zarr::S3Connection::S3Connection(const std::string& endpoint,
+                                 const std::string& access_key_id,
+                                 const std::string& secret_access_key,
+                                 const std::string& region)
```
jeskesen, line 27: This seems like unnecessary copy/paste.

A) could this be accomplished with with just a setter function?  Does it have to be in the constructor?  

B) If you still want a dedicated constructor could one constructor call the setter then call the other constructor? 

C) Alternatively have a region argument have default blank string that then doesn't set the region.

src/streaming/s3.connection.cpp
```diff
@@ -232,6 +248,27 @@ zarr::S3ConnectionPool::S3ConnectionPool(size_t n_connections,
     CHECK(!connections_.empty());
 }
 
+zarr::S3ConnectionPool::S3ConnectionPool(size_t n_connections,
```
jeskesen, line 251: Same comment as zarr::S3Connection::S3Connection

src/streaming/zarr.stream.cpp
```diff
@@ -546,11 +550,20 @@ ZarrStream_s::create_store_()
     if (is_s3_acquisition_()) {
         // spin up S3 connection pool
         try {
-            s3_connection_pool_ = std::make_shared<zarr::S3ConnectionPool>(
-              std::thread::hardware_concurrency(),
-              s3_settings_->endpoint,
-              s3_settings_->access_key_id,
-              s3_settings_->secret_access_key);
+            if (s3_settings_->region.has_value()) {
```
jeskesen, line 553: I feel like this would be cleaner if the S3ConnectionPool had a setter for the region, then only had that inside the region.has_value() conditional...  instead of the constructor with all its args.

jeskesen (approved):
I have some non-blocking style suggestions.  Even if you agree, they don't need to be done now, if this is urgent.

jeskesen (approved):

PR #60, opened 2025-02-13
Buffer into shards

aliddell:
- `ArrayWriter` subclasses now create a context-specific collection of buffers, depending on the type of array it is
  - `ZarrV2ArrayWriter` creates buffers for each *chunk*, `ZarrV3ArrayWriter` creates buffers for each *shard*; `ArrayWriter::make_buffers_()` is consequently made a pure virtual method
  - Implements a new virtual private method `ArrayWriter::get_chunk_data_()` which returns a pointer (either a full chunk buffer in the V2 case, or a segment of a shard buffer in the V3 case) that can be used in place of a vector
  - Implements a new private method `ArrayWriter::bytes_to_allocate_per_chunk_()` which returns either the canonical chunk size or the chunk size + a Blosc overhead, depending on whether we're compressing.
- Fixes a bug where multiscale frames do not get written on partial appends that fill the frame buffer.
- Tests writing data larger than one chunk in Python.
- Updates the benchmark program to run a single configuration, and provides a script to run them together.


nclack (changes requested):
@aliddell and I went through this together. 
- compressed chunk size array should get indexed by shard id
- this probably requires a slight refactoring of the compress_chunk function

nclack (approved):
lgtm. left one suggestion on a function name.

PR #66, opened 2025-02-21
Small fixes

### LINKED ISSUE #62: don't write acquire.zarr if there's no content
It looks like we write an "empty" acquire.json (It contains `{}`) when there's not `external_metadata`. 

Maybe we should just not write the file at all?

### LINKED ISSUE #64: Default clevel in Python bindings is 0
Should be at least 1.

aliddell:
- Closes #62 
- Closes #64


nclack (approved):
LGTM!

PR #68, opened 2025-02-25
Support OME-NGFF 0.5 in Zarr V3

aliddell:
The actual update is simple -- the multiscales metadata remains mostly the same, but is now under `attributes/ome` rather than just `attributes`. Also adds an overlooked `name` field to the `multiscales` metadata, which SHOULD be there by the spec.


jeskesen (approved):
We discussed verbally and all seems good.

Issue #55, opened 2025-02-06
asynchronous flush in scope?


tlambert03:
I've noticed that flushing to disk can block the calling thread rather significantly each time a chunk is filled (with the readme example of 3 channel 2k x 2k it blocks for around 600 ms on every flushing append (every 5 timepoints in my settings)

I cloned and tried rebuilding with `py::gil_scoped_release release;` on the `append()` function, but quickly realized that wasn't going to help since the call to `compress_and_flush_data` is itself synchronous, and [blocks until all jobs are done](https://github.com/acquire-project/acquire-zarr/blob/9a2bb904db8348a93922a9f01f223fac4cfda788/src/streaming/zarrv3.array.writer.cpp#L216).

I'm wondering if the discussion of asynchronous flushing has been discussed? For example, a double buffering approach where, once a buffer is full, it's swapped out and flushed asynchronously while a new buffer is immediately available for incoming data?  You all have much more experience that I do, so do you think that's just kicking the can down the road?  or might we expect to see some benefit?  (It is my _qualitative_ impression that the async write in tensorstore is allowing me to achieve this to some degree, but I have no numbers yet to prove that)

aliddell:
It's not a discussion we've actively had, at least not recently, but it's something I've been considering in order to improve performance. As you can imagine, memory use can get pretty large, but we wouldn't need to keep the second buffer around the entire time, just when flushing. I suspect this wouldn't take a lot of work. What do you think, @nclack?

tlambert03:
sounds good.

> As you can imagine, memory use can get pretty large, but we wouldn't need to keep the second buffer around the entire time

yeah.  

I'll try to write up what I feel to be a fair benchmark comparison.  would be good to have some solid sense for that, in a way that you feel is a proper usage.

nclack:
Double buffering sounds good to me. It's important to overlap appends and flushes.

tlambert03:
just had a chance to run this on a Windows computer... and the results were very different (with az being faster than tensorstore)

chunk size of (1, 2048, 2048)

```
uv run https://gist.githubusercontent.com/tlambert03/f8c1b069c2947b411ce24ea05aa370b1/raw/4dc107345370910e3c89c6f5cdc18759567fe11f/zarr_bench.py
```

```
Performance comparison:
  acquire-zarr: 525.583 ms
  TensorStore: 822.761 ms
  AZ/TS Ratio: 0.639
```

chunk size of (64, 512, 512)

```
[...]

PR #69, opened 2025-03-07
Create and destroy sinks on demand

aliddell:
Talley [noticed](https://github.com/acquire-project/acquire-zarr/issues/55#issuecomment-2676227165) an issue wherein Zarr datasets with small chunk sizes weren't writing all the data correctly. The reason turned out to be a situation where `FileSink`s would be created but not the underlying files not opened due to OS file handle exhaustion.

This PR fixes the issue by moving the creation of `Sink`s into the immediate context of where they're written, and enforces a limit (via semaphore) on the number of open handles allowed.

### DIFF
--- a/src/streaming/file.sink.hh
+++ b/src/streaming/file.sink.hh
@@ -9,7 +9,7 @@ namespace zarr {
 class FileSink : public Sink
 {
   public:
-    explicit FileSink(std::string_view filename);
+    FileSink(std::string_view filename, bool truncate = true);
 
     bool write(size_t offset, std::span<const std::byte> data) override;
 
--- a/src/streaming/zarrv2.array.writer.cpp
+++ b/src/streaming/zarrv2.array.writer.cpp
@@ -1,11 +1,14 @@
-#include "macros.hh"
 #include "zarrv2.array.writer.hh"
+
+#include "definitions.hh"
+#include "macros.hh"
 #include "sink.hh"
 #include "zarr.common.hh"
 
 #include <nlohmann/json.hpp>
 
 #include <latch>
+#include <semaphore>
 #include <stdexcept>
 
 namespace {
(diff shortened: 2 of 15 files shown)


src/streaming/file.sink.hh
```diff
@@ -9,7 +9,7 @@ namespace zarr {
 class FileSink : public Sink
 {
   public:
-    explicit FileSink(std::string_view filename);
+    FileSink(std::string_view filename, bool truncate = true);
```
jeskesen, line 12: Do you envision a scenario in which this would be false?   Many of the changes in this PR are dedicated to handling this boolean, but I never see it being set to anything but true. 

Also, you made the default true, and then also explicitly call true everywhere.   Not really a problem, just makes the PR touch much more than it needs to.

src/streaming/zarrv2.array.writer.cpp
```diff
@@ -117,64 +120,113 @@ zarr::ZarrV2ArrayWriter::get_chunk_data_(uint32_t index)
...
                         return false;
                     }
 
-                    const auto params = config_.compression_params;
-                    auto bytes_of_raw_chunk =
-                      config_.dimensions->bytes_per_chunk();
-
-                    auto* chunk_ptr = get_chunk_data_(i);
-
-                    try {
```
jeskesen, line 149: There are a few consecutive try/catch blocks that handle the same exception type similarly.  Might it be good to combine them?  Same comment for V3.

jeskesen (approved):

src/streaming/zarrv3.array.writer.cpp
```diff
@@ -306,66 +313,107 @@ zarr::ZarrV3ArrayWriter::compress_and_flush_data_()
...
+
+                         semaphore.acquire();
+                         if (s3_data_sinks_.contains(data_path)) {
+                             sink = std::move(s3_data_sinks_[data_path]);
+                         } else if (is_s3) {
+                             sink = make_s3_sink(
+                               *bucket_name, data_path, connection_pool);
+                         } else {
+                             sink =
+                               make_file_sink(data_path, *file_offset == 0);
```
aliddell, line 358: @jeskesen This is the one place where `truncate` might be false.

PR #71, opened 2025-03-13
Explicitly assign S3 port when none is specified

### LINKED ISSUE #70: S3 default port not set correctly
When defining the endpoint in an S3 acquisition, if the user doesn't specify the port value, minio-cpp does not define it to be port 80, as you might expect. The value is either unset or set to 10.

aliddell:
closes #70

### DIFF
--- a/src/streaming/zarr.stream.cpp
+++ b/src/streaming/zarr.stream.cpp
@@ -543,20 +543,8 @@ ZarrStream_s::create_store_()
     if (is_s3_acquisition_()) {
         // spin up S3 connection pool
         try {
-            if (s3_settings_->region.has_value()) {
-                s3_connection_pool_ = std::make_shared<zarr::S3ConnectionPool>(
-                  std::thread::hardware_concurrency(),
-                  s3_settings_->endpoint,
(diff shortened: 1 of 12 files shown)


src/streaming/zarr.stream.cpp
```diff
@@ -543,20 +543,8 @@ ZarrStream_s::create_store_()
...
-                  *s3_settings_->region);
-            } else {
-                s3_connection_pool_ = std::make_shared<zarr::S3ConnectionPool>(
-                  std::thread::hardware_concurrency(),
-                  s3_settings_->endpoint,
-                  s3_settings_->access_key_id,
-                  s3_settings_->secret_access_key);
-            }
+            s3_connection_pool_ = std::make_shared<zarr::S3ConnectionPool>(
+              std::thread::hardware_concurrency(), *s3_settings_);
```
nclack, line 547: shouldn't this follow the thread limit that a user sets?

If so, that doesn't have to get fixed in this PR; just create an issue.

nclack (approved):
lgtm! asked for a comment and had a question.

PR #72, opened 2025-03-20
Some performance enhancements

aliddell:
- Uses platform-specific file I/O. Before the other interventions, I found this modestly improved performance of file I/O on all machines I tested on.
- Precomputes several values which are often queried from `ArrayDimension`.
- Uses OpenMP and wraps the main `for` loop in `ArrayWriter::write_frame_to_chunks_()` in a `#pragma omp parallel for`.


src/streaming/array.writer.cpp
```diff
@@ -224,44 +225,42 @@ zarr::ArrayWriter::write_frame_to_chunks_(std::span<const std::byte> data)
...
-                    bytes_written += (region_stop - region_start);
-                }
-                chunk_ptr += static_cast<long long>(bytes_per_row);
+    const auto* data_ptr = data.data();
+    const auto data_size = data.size();
+
+    size_t bytes_written = 0;
+    const auto n_tiles = n_tiles_x * n_tiles_y;
+
+#pragma omp parallel for reduction(+ : bytes_written)
```
nclack, line 234: Check this with a benchmark, but this will scale better if you make bytes_written an atomic and a plain old parallel for.

Most of the work here is not involved with the reduction, and the way reductions get parallelized might be limiting.

Maybe omp figures this out for you, but it's better not to rely on that.

src/streaming/file.sink.cpp
```diff
@@ -3,31 +3,226 @@
...
+#include <unistd.h>
+#endif
+
 namespace fs = std::filesystem;
 
-zarr::FileSink::FileSink(std::string_view filename, bool truncate)
-  : file_(filename.data(),
-          truncate ? (std::ios::binary | std::ios::trunc) : std::ios::binary)
+namespace {
+#ifdef _WIN32
```
nclack, line 18: If there's a way to factor out the platform-dependent things into their own source files, it makes the code much easier to read and maintain.

src/streaming/zarr.stream.cpp
```diff
@@ -7,13 +7,18 @@
...
 #include <blosc.h>
+#include <omp.h>
 
 #include <filesystem>
 
 #ifdef min
 #undef min
 #endif
 
+#ifdef max
```
nclack, line 18: You might be able to just say #undef without needing the #ifdef.

More importantly, where are these symbols leaking in from? If possible you should handle this as close to the source as possible.

nclack (commented):
Left some comments. Be sure to break the performance improvements into multiple pr's if you can.

jeskesen (dismissed):

src/streaming/array.writer.cpp
aliddell: I actually see significant improvement with the reduction.

- atomic: 1.081 GiB/s
- reduction: 1.354 GiB/s

I ran multiple times and saw similar numbers.

jeskesen (commented):
Trying to remove my "approve", because I hit it accidentally before.

src/streaming/file.sink.cpp
```diff
@@ -3,31 +3,226 @@
...
+
+    size_t size = FormatMessageA(FORMAT_MESSAGE_ALLOCATE_BUFFER |
+                                   FORMAT_MESSAGE_FROM_SYSTEM |
+                                   FORMAT_MESSAGE_IGNORE_INSERTS,
+                                 NULL,
+                                 errorMessageID,
+                                 MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+                                 (LPSTR)&messageBuffer,
+                                 0,
+                                 NULL);
```
shlomnissan, line 37:
```cpp
const auto lang = MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT);
const auto format = FORMAT_MESSAGE_ALLOCATE_BUFFER |
                    FORMAT_MESSAGE_FROM_SYSTEM |
                    FORMAT_MESSAGE_IGNORE_INSERTS;

auto buffer = static_cast<LPSTR*>(&message_buffer);
DWORD size = FormatMessageA(format, nullptr, error_message_id, lang, 0, buffer, nullptr);
```

- Replace `NULL` with `nullptr`.
- Ensure consistent naming e.g., `error_message_id` all lowercase.
- Use `static_cast` explicitly for the buffer if possible.

src/streaming/file.sink.cpp
```diff
@@ -3,31 +3,226 @@
...
+
+    std::string message(messageBuffer, size);
+
+    LocalFree(messageBuffer);
+
+    return message;
+}
+
+void
+init_handle(void** handle, const std::string& filename)
```
shlomnissan, line 47: The `void**` pattern is error-prone and hard to reason about. Is there a reason not to use a `unique_ptr` or a small RAII wrapper instead?

src/streaming/file.sink.cpp
```diff
@@ -3,31 +3,226 @@
...
+                      GENERIC_WRITE,
+                      0, // No sharing
+                      nullptr,
+                      OPEN_ALWAYS,
+                      FILE_FLAG_OVERLAPPED,
+                      nullptr);
+
+    if (*fd == INVALID_HANDLE_VALUE) {
+        const auto err = get_last_error_as_string();
+        delete handle;
```
shlomnissan, line 62: I _think_ this is a bug?
```suggestion
        delete fd;
```

src/streaming/file.sink.cpp
```diff
@@ -3,31 +3,226 @@
...
+        throw std::runtime_error("Failed to open file '" + filename +
```
shlomnissan, line 63: I think you can just use `std::format` now:
```cpp
throw std::runtime_error(std::format("Failed to open file {}: {}", filename, err);
```
This applies to other string formatting around errors in this PR.

src/streaming/file.sink.cpp
```diff
@@ -3,31 +3,226 @@
...
+}
+
+bool
+seek_and_write(void** handle, size_t offset, ConstByteSpan data)
+{
+    CHECK(handle);
+    auto* fd = reinterpret_cast<HANDLE*>(*handle);
+
+    auto* cur = reinterpret_cast<const char*>(data.data());
+    auto* end = cur + data.size();
```
shlomnissan, line 76: If you are already casting to a pointer or assigning a pointer, I don't think adding a `*` is necessary:

```suggestion
    auto fd  = reinterpret_cast<HANDLE*>(*handle);
    auto curr = reinterpret_cast<const char*>(data.data());
    auto end = cur + data.size();
```

This applies to other places in this PR.

src/streaming/file.sink.cpp
```diff
@@ -3,31 +3,226 @@
...
+
+    int retries = 0;
+    OVERLAPPED overlapped = { 0 };
+    overlapped.hEvent = CreateEventA(nullptr, TRUE, FALSE, nullptr);
```
shlomnissan, line 80: I think that if the function expects integers, you should pass integers instead of `TRUE`/`FALSE` macros, otherwise, always prefer `true`/`false` keywords.

src/streaming/file.sink.cpp
```diff
@@ -3,31 +3,226 @@
...
+
+    while (cur < end && retries < 3) {
```
shlomnissan, line 82: Avoid magic numbers. I think creating a variable `max_retries` is more robust and easier to read. This is also hard-coded in other areas in the code so a named variable would be a much better approach.

src/streaming/file.sink.cpp
```diff
@@ -3,31 +3,226 @@
...
+        DWORD written = 0;
+        DWORD remaining = (DWORD)(end - cur); // may truncate
```
shlomnissan, line 84: Prefer static casting (i.e., `static_cast<DWORD>(end - cur)`) to C-style casting, for type safety, readability, etc. There are several places in this PR where this is applicable.

src/streaming/file.sink.cpp
```diff
@@ -3,31 +3,226 @@
...
+            return false;
+        }
+
+        if (!GetOverlappedResult(*fd, &overlapped, &written, TRUE)) {
+            LOG_ERROR("Failed to get overlapped result: ",
+                      get_last_error_as_string());
+            CloseHandle(overlapped.hEvent);
+            return false;
+        }
+        retries += (written == 0);
```
shlomnissan, line 100: Try to avoid implicit casting if possible.
```suggestion
        retries += (written == 0) ? 1 : 0;
```

src/streaming/file.sink.cpp
```diff
@@ -3,31 +3,226 @@
...
+}
+#else
+std::string
+get_last_error_as_string()
+{
+    return strerror(errno);
+}
+
+void
+init_handle(void** handle, const std::string& filename)
```
shlomnissan, line 135: I think you could use `std::string_view` instead to increase the range of this function.

src/streaming/file.sink.cpp
shlomnissan: 100%

src/streaming/zarr.dimension.cpp
```diff
@@ -6,8 +6,34 @@ ArrayDimensions::ArrayDimensions(std::vector<ZarrDimension>&& dims,
...
+    }
+
+    for (const auto& dim : dims_) {
+        chunks_per_shard_ *= dim.shard_size_chunks;
+    }
+
+    for (auto i = 1; i < ndims(); ++i) {
+        const auto& dim = dims_[i];
+        number_of_shards_ *= zarr::shards_along_dimension(dim);
+    }
```
shlomnissan, line 31: I think that you can collapse all of these loops into one unless I'm missing something?

```cpp
for (size_t i = 0; i < dims_.size(); ++i) {
    const auto& dim = dims_[i];
    bytes_per_chunk_ *= dim.chunk_size_px;
    chunks_per_shard_ *= dim.shard_size_chunks;
    if (i > 0) {
        number_of_chunks_in_memory_ *= zarr::chunks_along_dimension(dim);
        number_of_shards_ *= zarr::shards_along_dimension(dim);
    }
}
```

shlomnissan (changes requested):
I’ve flagged one potential bug — it looks like the wrong pointer might be getting deleted, so I’m requesting changes just to ensure that’s double-checked.

Beyond that, I think there’s a lot of room to improve the robustness and readability of the code. The manual pointer management, in particular, feels risky and could lead to maintenance issues. If possible, I'd recommend avoiding raw pointers in favor of safer patterns.

That said, the rest of my comments are just suggestions — feel free to weigh them against your judgment.

.github/workflows/build.yml
```diff
@@ -38,13 +44,18 @@ jobs:
 
       - name: Install vcpkg
         run: |
-          git clone https://github.com/microsoft/vcpkg.git
+          git clone https://github.com/microsoft/vcpkg.git -b 2025.02.14
```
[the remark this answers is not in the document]
aliddell: I ran into an issue in CI on Friday afternoon which was introduced with [this](https://github.com/microsoft/vcpkg/pull/44537) and apparently resolved with [this](https://github.com/microsoft/vcpkg/pull/44543). I decided to tag a known working version so this doesn't happen again.

src/streaming/file.sink.cpp
aliddell: Here is the function signature:

```c++
WINBASEAPI
_Ret_maybenull_
HANDLE
WINAPI
CreateEventA(
    _In_opt_ LPSECURITY_ATTRIBUTES lpEventAttributes,
    _In_ BOOL bManualReset,
    _In_ BOOL bInitialState,
    _In_opt_ LPCSTR lpName
    );
```

The `BOOL` is `typedef int BOOL`, so while technically it is expecting integers, I'm calling the function as it's designed to be called.

src/streaming/array.writer.cpp
```diff
@@ -224,44 +224,43 @@ zarr::ArrayWriter::write_frame_to_chunks_(std::span<const std::byte> data)
...
-                              data.begin() + region_stop,
-                              chunk_ptr);
-
-                    bytes_written += (region_stop - region_start);
-                }
-                chunk_ptr += static_cast<long long>(bytes_per_row);
+    const auto* data_ptr = data.data();
+    const auto data_size = data.size();
+
+    size_t bytes_written = 0;
```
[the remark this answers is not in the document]
aliddell: I read this page and it doesn't seem to support what you're suggesting. I'm not using `auto` to avoid uninitialized variables or type mismatches, and I'm not trying to deduce an "invisible proxy class" (like `std::vector<bool>::reference`. Plus it's needlessly verbose.

cmake/macos-toolchain.cmake
```diff
@@ -0,0 +1,31 @@
...
+        set(LIBOMP_PATH "/opt/homebrew/opt/libomp")
+        set(CMAKE_OSX_ARCHITECTURES "arm64")
+    else ()
+        # Intel Mac
+        set(LIBOMP_PATH "/usr/local/opt/libomp")
+        set(CMAKE_OSX_ARCHITECTURES "x86_64")
+    endif ()
+
+    # OpenMP support
+    set(OpenMP_C_FLAGS "-Xclang -fopenmp -I${LIBOMP_PATH}/include")
```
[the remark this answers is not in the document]
aliddell: Not at this point, but in case any is added down the line I don't want the configuration to break inexplicably for future users.

cmake/macos-toolchain.cmake
```diff
@@ -0,0 +1,31 @@
+if (APPLE)
```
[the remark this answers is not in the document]
aliddell: Good idea.

aliddell:
Thanks for comments @nclack @shlomnissan. I've made the following changes:

- Renamed `macos-toolchain.cmake` to `openmp.cmake`.
- Added instructions to the README for Mac users to install OpenMP with Homebrew.
- Broken out Windows and POSIX implementations for file.sink.cpp.
- Added a comment above `#pragma omp parallel for` describing why we only use 75% of the thread pool allotment for OMP parallelism. (Will make an issue for this once it's merged.)
- Updated the interior of the parfor loop to have some more descriptive variable names.
  - Added some more verbiage to help with debugging in the case of buffer overrun.
- Used `auto` and `static_cast<T>` a bit more liberally in the file I/O code.
- **Fixed the bug Shlomi found re: deleting the wrong pointer**
- Consolidated a few `for` loops in the `ArrayDimension` constructor.

shlomnissan (approved):
Unblocking now that the bug is fixed.

PR #73, opened 2025-03-26
Implement and test a frame queue

aliddell:
Implements and tests and simple lock-free frame queue.

Also, we don't use `VectorizedFileWriter`, so I removed it.


jeskesen (commented):
So, you deleted vectorized file writer because it wasn't used, but neither is FrameQueue.  Or will that come at a later stage?

aliddell:
> So, you deleted vectorized file writer because it wasn't used, but neither is FrameQueue. Or will that come at a later stage?

It comes later. See #75

nclack (commented):
It's a little hard to evaluate this without looking at #75. I need to know the requirements for FrameQueue. I'm not sure if it needs to be thread safe, for example. Can you clarify in the PR description?

PR #80, opened 2025-04-02
Add Talley's benchmark comparing acquire-zarr to TensorStore

aliddell:


nclack (approved):

nclack (approved):
Can you make sure the version with the per-frame append times gets in?

tlambert03 (approved):
looks good to me.  couple minor (non-blocking) notes:

1. The data comparison seems _much_ slower the original method...  I haven't looked closely enough to understand the need for the CyclicArray structure and the piece-wise data comparison (I assume there's a good reason).
2. Just a side-note... I do think it's possible that I'm not using tensorstore as optimally as I possibly could be.  I asked a quick question to the tensorstore folks and they highlighted that I could be using [transactions](https://google.github.io/tensorstore/python/api/tensorstore.Transaction.html) rather than making my own numpy chunk-cache object.  So it's possible that would speed things up (or at least make the TS code cleaner).  Will let you know if I learn more about that implementation

PR #88, opened 2025-04-08
Downsample in 2d or 3d depending on third dimension

aliddell:
Automatically determines whether or not downsampling is 2- or 3-dimensional and downsamples accordingly. If the third (i.e., the dimension after width and height) dimension is spatial, downsampling is 3-dimensional and consecutive frames are averaged (this is the old behavior). Otherwise, we only downsample width and height.

Minor refactor -- puts downsampling in a class of its own.


src/streaming/downsampler.cpp
```diff
@@ -0,0 +1,295 @@
...
+    const auto bytes_of_frame = width * height * sizeof(T);
+
+    EXPECT(bytes_of_src >= bytes_of_frame,
+           "Expecting at least ",
+           bytes_of_frame,
+           " bytes, got ",
+           bytes_of_src);
+
+    const int downscale = 2;
+    constexpr auto bytes_of_type = static_cast<double>(sizeof(T));
```
nclack, line 21: use of floating point throughout this function is wrong. The calculations here around dimensions should be fixed-point to be correct. Also, prefer fixed width types (int32_t vs int, (u)int64_t vs size_t) where possible. This will make it clearer where you need a certain precision.

src/streaming/downsampler.cpp
```diff
@@ -0,0 +1,295 @@
...
+    const double factor = 0.25;
+
+    const auto w_pad = static_cast<double>(width + (width % downscale));
```
nclack, line 24: Use of `auto` is not making this code clearer. The many uses of static_cast and reinterpret_cast also look like something is wrong. Cast to what you want at the top of the function/outside the loops so things are already in the right types.

src/streaming/downsampler.cpp
```diff
@@ -0,0 +1,295 @@
...
+    for (auto row = 0; row < height; row += downscale) {
+        const bool pad_height = (row == height - 1 && height != h_pad);
+
+        for (auto col = 0; col < width; col += downscale) {
+            size_t src_idx = row * width + col;
+            const bool pad_width = (col == width - 1 && width != w_pad);
+
+            auto here = static_cast<double>(src_as_T[src_idx]);
+            auto right = static_cast<double>(
+              src_as_T[src_idx + (1 - static_cast<int>(pad_width))]);
```
nclack, line 44: you can precompute these offsets outside of the for loop

src/streaming/downsampler.cpp
```diff
@@ -0,0 +1,295 @@
...
+            auto down = static_cast<double>(
+              src_as_T[src_idx + width * (1 - static_cast<int>(pad_height))]);
+            auto diag = static_cast<double>(
+              src_as_T[src_idx + width * (1 - static_cast<int>(pad_height)) +
+                       (1 - static_cast<int>(pad_width))]);
+
+            dst_as_T[dst_idx++] =
+              static_cast<T>(factor * (here + right + down + diag));
```
nclack, line 52: this is one of those cases where division might work better, since it avoids a cast to float for fixed point types.

`T(here+right+down+diag)/T(4)`

src/streaming/downsampler.cpp
```diff
@@ -0,0 +1,295 @@
...
+            const bool pad_width = (col == width - 1 && width != w_pad);
+
+            auto here = static_cast<double>(src_as_T[src_idx]);
+            auto right = static_cast<double>(
+              src_as_T[src_idx + (1 - static_cast<int>(pad_width))]);
+            auto down = static_cast<double>(
+              src_as_T[src_idx + width * (1 - static_cast<int>(pad_height))]);
+            auto diag = static_cast<double>(
+              src_as_T[src_idx + width * (1 - static_cast<int>(pad_height)) +
+                       (1 - static_cast<int>(pad_width))]);
```
nclack, line 49: I don't understand some parts of the padding/offset calculation entirely at the moment, though I'm sure this can work.

I think it's fine just to do boundary clamping inside the loop.

`row1 = max(row+1,height-1)` etc.

Alternatively, could also make a decision just to truncate the odd edges. So if you had a 3x3 input, the output is 1x1 - or if you had 513x512 the output is 256x256.

We need to document the decision and test against it since it's an important detail.

mtcaton:
i ran my script with this version of the code and it seemed to work. the zarr.json file at the group level was fixed and i was able to view the image in napari. @nclack do you think your comments need to be addressed before I fully ingest the data?

src/streaming/downsampler.cpp
aliddell: Now that I think about this more, I don't know if this is even the correct approach. Consider a 2x2 grid with values 1, 1, 1, 0. The way I do it and the way you suggest would both truncated the averaged value to 0, which means we'd lose the dominant signal. Should we be rounding instead? We could do

```
dst_as_T[dst_idx++] = static_cast<T>((here + right + down + diag + 2) / 4);
```

which would effectively round to the nearest integer.

src/streaming/downsampler.cpp
aliddell: This is one of those "too-clever-by-half" things I wrote. For example, `pad_width` is `true` iff `col` is odd and equal to `width - 1`, so if we're padding (because we're on the edge) it'll just take the same pixel twice and if we're not, it'll take the pixel adjacent to it.

Not sure what you're describing when you say boundary clamping inside the loop.

nclack (approved):

PR #90, opened 2025-04-10
Fix transposed array bug

aliddell:
Shallowly transposing image arrays (as in `np.transpose()`) and passing them to `stream.append()` fails to write the data in the order the user expects. This PR fixes that by making a contiguous copy, if necessary, and passing that data to `stream.append()`.

### DIFF
--- a/python/acquire-zarr-py.cpp
+++ b/python/acquire-zarr-py.cpp
@@ -109,6 +109,8 @@ log_level_to_str(ZarrLogLevel level)
             return "UNKNOWN";
     }
 }
+
+py::module np;
 } // namespace
 
 class PyZarrS3Settings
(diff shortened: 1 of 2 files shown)


python/acquire-zarr-py.cpp
```diff
@@ -393,7 +393,16 @@ class PyZarrStream
...
 
-        auto buf = image_data.request();
+        // Create a contiguous copy, but only if needed
+        py::array contiguous_data;
+        if (!(image_data.flags() & py::array::c_style)) {
+            py::module np = py::module::import("numpy");
+            contiguous_data = np.attr("ascontiguousarray")(image_data);
+        } else {
+            contiguous_data = image_data;
+        }
```
jeskesen, line 403: from what I read, using the function np.ascontiguiousarray() performs the conditional check for you, and only copies if necessary.

Also, its probably better to move the numpy import to the beginning, to front-load that call (and spot failure earlier?)

jeskesen (changes requested):

jeskesen (commented):
I'll be honest, I was hoping this would lead to a more general "writing out-of-order data" solution that could also be used in C++, but this fixes the issue at hand.

jeskesen (approved):
LGTM.

jeskesen:
@aliddell , I merged it, but didn't squash or delete the branch, so one could continue working on the branch.

nclack:
What would happen if we altered the dimension spec instead?

PR #98, opened 2025-04-14
Fix handling of transposed NumPy arrays in `ZarrStream.append()`

aliddell:
This PR addresses an issue where transposed NumPy arrays would be written incorrectly when passed directly to `append()`. The problem occurs because NumPy's transpose operation creates a view with modified strides rather than rearranging the data in memory, but our C++ code was interpreting the memory linearly. We had previously made a [copy](https://github.com/acquire-project/acquire-zarr/pull/90), assuming data would be small, but of course we have no such guarantee.

## Changes

- Added logic to detect non-contiguous arrays (like those created by `np.transpose()`)
- For smaller, low-dimensional arrays (1D/2D), create a contiguous copy before processing
- For larger multi-dimensional arrays, iterate through frames to conserve memory
- Process each frame individually, ensuring correct memory layout

### DIFF
--- a/python/acquire-zarr-py.cpp
+++ b/python/acquire-zarr-py.cpp
@@ -367,13 +367,79 @@ class PyZarrStream
             throw py::error_already_set();
         }
 
-        // Create a contiguous copy, but only if needed
-        py::array contiguous_data;
-        if (!(image_data.flags() & py::array::c_style)) {
+        // if the array is already contiguous, we can just write it out
+        if (image_data.flags() & py::array::c_style) {
+            write_contiguous_data(image_data);
+            return;
+        }
+
+        // just make a copy of smaller (2-dim or less) arrays
+        if (image_data.ndim() <= 2) {
             py::module np = py::module::import("numpy");
-            contiguous_data = np.attr("ascontiguousarray")(image_data);
+            py::array contiguous_data =
+              np.attr("ascontiguousarray")(image_data);
+            write_contiguous_data(contiguous_data);
+            return;
+        }
+
+        // iterate through frames
+        iterate_and_append(image_data, 0, std::vector<py::ssize_t>());
+    }
+
+    // iterate over the indices of the array until we get down to 2 dimensions,
+    // then write the frame
+    void iterate_and_append(const py::array& array,
+                            size_t dim,
+                            std::vector<py::ssize_t> indices)
+    {
+        if (dim == array.ndim() - 2) {
+            // we are down to a 2D frame - we can write it
+            py::array frame = extract_frame(array, indices);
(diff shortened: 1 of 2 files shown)


python/acquire-zarr-py.cpp
```diff
@@ -367,23 +367,89 @@ class PyZarrStream
...
+        iterate_and_append(image_data, 0, std::vector<py::ssize_t>());
+    }
+
+    // iterate over the indices of the array until we get down to 2 dimensions,
+    // then write the frame
+    void iterate_and_append(const py::array& array,
+                            size_t dim,
+                            std::vector<py::ssize_t> indices)
+    {
+        if (dim == array.ndim() - 2) {
```
jeskesen, line 395: Are you actually down to a 2D array here, or an ndim()-2 sized array?

python/acquire-zarr-py.cpp
```diff
@@ -367,23 +367,89 @@ class PyZarrStream
...
+            // we are down to a 2D frame - we can write it
+            py::array frame = extract_frame(array, indices);
+            write_contiguous_data(frame);
```
jeskesen, line 398: should you also do the "ascontiguousarray" here?

python/acquire-zarr-py.cpp
```diff
@@ -367,23 +367,89 @@ class PyZarrStream
...
+        }
+
+        // just make a copy of smaller (2-dim or less) arrays
+        if (image_data.ndim() <= 2) {
+            py::module np = py::module::import("numpy");
+            py::array contiguous_data =
+              np.attr("ascontiguousarray")(image_data);
+            write_contiguous_data(contiguous_data);
+            return;
+        }
```
jeskesen, line 383: does the n<=2 case need to be called out explicitly, or wouldn't calling iterate_and_append also work (after a single iteration).

python/acquire-zarr-py.cpp
aliddell: No, it builds up the indices starting with the outermost to the innermost. So if we have a TCZYX array, it'll do T (dim = 1), then TC (dim = 2), then TCZ (dim = 3 = ndim() - 2).

python/acquire-zarr-py.cpp
jeskesen: But that would also be handled in extract_frame by numpy, as you pointed out in a reply to a comment below.

aliddell: I mean, yes, I could construct an array of dimensions and pass it into `extract_frame`, or I could just handle the simple case directly.

jeskesen (approved):

PR #99, opened 2025-04-17
Fix regression involving column-ragged shards

aliddell:
## Problem
PR #60 introduced a regression in writing Zarr V3 arrays where specifically shard columns are ragged. In `ZarrV3ArrayWriter::get_chunk_data_()`, the method tacitly assumes that the shard is not sparse, using the "shard internal index" of each chunk as a shorthand for where in the shard buffer the chunk data is stored. However, when the shard is written to disk, only the actual number of chunks' worth of shard data is written.

This creates a mismatch between the location where chunk data is stored in memory and the location where it should be in the final output, specifically when:

- Dimensions don't divide evenly into chunks
- The resulting shard has an irregular number of chunks
- The internal index calculation doesn't align with the actual chunk position

## Solution

This PR fixes the issue by:

- Adding a new mapping (`chunk_indices_for_shard_`) that tracks which chunk indices belong to each shard
- Modifying `get_chunk_data_()` to check if the calculated internal index corresponds to the expected chunk index
- If they don't match (indicating a ragged shard), finding the correct position by looking up the chunk index in the shard's chunk list
- Using this corrected offset to place the chunk data at the proper location in memory

The patch includes a test case specifically designed to verify correct handling of column-ragged shards, which should catch any future regressions.

## Technical Details

- In `ArrayDimensions`, we now maintain a vector of chunk indices for each shard
- When initializing dimensions, we populate both the existing map (chunk -> shard) and the new map (shard -> chunks)
- In the writer, we verify the internal index matches the expected chunk and adjust as needed
- This ensures consistent behavior regardless of whether the shard structure is regular or ragged

This change maintains backward compatibility with existing code while fixing the specific edge case of column-ragged shards.


nclack (commented):
Most of the memory you're allocating here is used for the internal chunks where the chunk_index_for_shard calculation can be trivial. I believe there's a way to do this without allocating any memory in O(shard_size_chunks in x) time.

Am I right that this is a 2d problem only?

So first step: think about how to implement `dims->shard_internal_index(index)` by writing a for loop where you basically count the in-bound chunks that hit the shard before `index`.

Second step: that for-loop can get a lot more efficient if we only iterate over the chunks in the shard. We know the index of the first chunk for the shard, the stride (`n_tiles_x`). This is a 2d for loop over the shard dimensions.

Last: The chunk rows are all ragged in basically exactly the same way. So you can calculate how many are in-bounds in a single chunk row and then multiply by the number of rows before index.

jeskesen:
> Most of the memory you're allocating here is used for the internal chunks where the chunk_index_for_shard calculation can be trivial. I believe there's a way to do this without allocating any memory in O(shard_size_chunks in x) time.
> 
> Am I right that this is a 2d problem only?
> 
> So first step: think about how to implement `dims->shard_internal_index(index)` by writing a for loop where you basically count the in-bound chunks that hit the shard before `index`.
> 
> Second step: that for-loop can get a lot more efficient if we only iterate over the chunks in the shard. We know the index of the first chunk for the shard, the stride (`n_tiles_x`). This is a 2d for loop over the shard dimensions.
> 
> Last: The chunk rows are all ragged in basically exactly the same way. So you can calculate how many are in-bounds in a single chunk row and then multiply by the number of rows before index.



> Most of the memory you're allocating here is used for the internal chunks where the chunk_index_for_shard calculation can be trivial. I believe there's a way to do this without allocating any memory in O(shard_size_chunks in x) time.
> 
> Am I right that this is a 2d problem only?
> 
> So first step: think about how to implement `dims->shard_internal_index(index)` by writing a for loop where you basically count the in-bound chunks that hit the shard before `index`.
> 
> Second step: that for-loop can get a lot more efficient if we only iterate over the chunks in the shard. We know the index of the first chunk for the shard, the stride (`n_tiles_x`). This is a 2d for loop over the shard dimensions.
> 
> Last: The chunk rows are all ragged in basically exactly the same way. So you can calculate how many are in-bounds in a single chunk row and then multiply by the number of rows before index.

@nclack , this seems like a lot of optimization for something that I think is linear for a small N.  Also, does the allocation happen every frame, or just once up front? (I thought it was the former)

nclack:
Alan convinced me the lookup table is fast. We also discussed `group_offset` handles ragedness in higher dimensions.

nclack (approved):

PR #100, opened 2025-04-18
Automatically detect trivial Z dimension when downsampling

aliddell:
This PR addresses two issues in the multiscale downsampling implementation:

1. Fixed 3D downsampling detection to only apply when the third spatial dimension is
   actually non-trivial (more than 1 element). Previously, we would attempt 3D downsampling
   even for essentially 2-spatial-D datasets that had a spatial Z dimension with only one plane.

2. Revised the multiscale coordinate transform calculation to use actual dimension ratios
   instead of fixed power-of-2 scaling factors. This ensures that the OME-NGFF metadata
   accurately reflects the true relationship between resolution levels for all dimensions.

### DIFF
--- a/src/streaming/downsampler.cpp
+++ b/src/streaming/downsampler.cpp
@@ -158,11 +158,13 @@ bool
 zarr::Downsampler::is_3d_downsample_() const
 {
     // the width and depth dimensions are always spatial -- if the 3rd dimension
-    // is also spatial, then we downsample in 3 dimensions
+    // is also spatial and nontrivial, then we downsample in 3 dimensions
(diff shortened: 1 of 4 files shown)


jeskesen (approved):
I'm not sure I 100% understand the scaling calculation, but I don't want to hold it up...

Other than that, LGTM.

aliddell:
@jeskesen it rounds to the next power of 2. I'm sure there's a better way to do it. I'll add a comment.

jeskesen (commented):
@aliddell I realize this is already merged, but after a little googling, I found this:  https://en.cppreference.com/w/cpp/numeric/bit_ceil

PR #105, opened 2025-05-08
Refactor array/group to node

aliddell:
This PR introduces a class `Node`, which encompasses both a Zarr group and a Zarr array. This PR also introduces a `Group` class which contains one or more `Array`s. `ArrayWriter` and its subclasses have been named to `Array` (and `V2Array` and `V3Array`, respectively. The Zarr stream is no longer responsible for writing metadata, but instead now maintains a member, `output_node_`, which is an instance of (pointer to) `Node` which writes its own metadata.

This PR is in anticipation of a future PR which allows users to specify a node within a dataset as the destination for an output stream, which may be either a multiscale group or simply an array.

There are no user-facing changes.


src/streaming/node.hh
```diff
@@ -0,0 +1,97 @@
...
+#include "definitions.hh"
+#include "s3.connection.hh"
+#include "sink.hh"
+#include "thread.pool.hh"
+#include "zarr.types.h"
+
+#include <string>
+
+namespace zarr {
+struct NodeConfig
```
nclack, line 14: The name `Node` is too generic. Some possible alternatives 'ZarrElement', 'ZarrContainer'.

src/streaming/node.hh
```diff
@@ -0,0 +1,97 @@
...
+
+    std::string store_root;
+    std::string group_key;
+    std::optional<std::string> bucket_name;
+    std::optional<BloscCompressionParams> compression_params;
+    std::shared_ptr<ArrayDimensions> dimensions;
+    ZarrDataType dtype;
+};
+
+class Node
```
nclack, line 42: You're using inheritance a lot here, which will ultimately make the code harder to reason about. It's better to prefer a compositional style e.g. make IFrameWriter and IMetadataWriter interface classes and compose implementations of those to make the generic Group or Arrays.

That said, I won't block the PR for this.

src/streaming/array.cpp
```diff
@@ -140,60 +67,74 @@ zarr::ArrayWriter::write_frame(std::span<const std::byte> data)
     return bytes_written;
 }
 
+bool
+zarr::Array::close_()
+{
+    bool retval = false;
+    is_closing_ = true;
```
nclack, line 74: `is_closing_` looks like it's only read in `v3.array.cpp`. Is that right? 

I'm sort of worried about a race condition on is_closing_ in compress_and_flush

src/streaming/node.cpp
```diff
@@ -0,0 +1,107 @@
...
+            LOG_ERROR("Failed to write metadata for key: ", key);
+            return false;
+        }
+    }
+
+    return true;
+}
+
+bool
+zarr::finalize_node(std::unique_ptr<Node>&& node)
```
nclack, line 81: This looks super fragile. Why should Node have to know how to delete its child classes? Shouldn't an output_node know how to delete itself properly?

nclack (commented):
I left some comments. Mostly, I'd just like you to find a better name for the Node class.

src/streaming/node.hh
jeskesen: said like a man who prefers C to C++.  ;)

src/streaming/array.cpp
```diff
@@ -1,107 +1,34 @@
...
-        const auto& dim = config.dimensions->at(i);
-        const auto& downsampled_dim = downsampled_config.dimensions->at(i);
-
-        if (dim.chunk_size_px > downsampled_dim.chunk_size_px) {
-            return false;
-        }
-    }
-
-    return true;
-}
```
aliddell, line 76: Dead code made unnecessary by #88.

nclack (approved):

PR #108, opened 2025-05-13
Expose method used to downsample arrays

### LINKED ISSUE #101: Expose downsampling function for creating multi-scale arrays
For label arrays, downsampling should be something other than the current hard-coded `downscale_local_mean` to avoid interpolation.

Minimal solution would be simple slicing (=subsampling).

Mode pooling would be nicer, but this might be quite slow.

Passing an arbitrary function would be cool, but not sure whether this is a desirable design choice.

aliddell:
Closes #101.

### Methods supported

- Mean
- Min
- Max

## API changes

A new parameter `downsampling_method` was added to the `StreamSettings` struct (`ZarrStreamSettings` class in Python).

### DIFF
--- a/python/acquire-zarr-py.cpp
+++ b/python/acquire-zarr-py.cpp
@@ -272,6 +272,15 @@ class PyZarrStreamSettings
         max_threads_ = max_threads;
     }
 
+    ZarrDownsamplingMethod downsampling_method() const
+    {
+        return downsampling_method_;
+    }
+    void set_downsampling_method(ZarrDownsamplingMethod method)
+    {
+        downsampling_method_ = method;
+    }
+
     const std::vector<PyZarrDimensionProperties>& dimensions() const
     {
         return dimensions_;
@@ -294,6 +303,7 @@ class PyZarrStreamSettings
     ZarrVersion version_{ ZarrVersion_2 };
     unsigned int max_threads_{ std::thread::hardware_concurrency() };
     std::vector<PyZarrDimensionProperties> dimensions_;
+    ZarrDownsamplingMethod downsampling_method_{ ZarrDownsamplingMethod_None };
 };
 
 class PyZarrStream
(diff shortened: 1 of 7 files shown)


python/acquire-zarr-py.cpp
```diff
@@ -588,6 +599,12 @@ PYBIND11_MODULE(acquire_zarr, m)
       .value(dimension_type_to_str(ZarrDimensionType_Other),
              ZarrDimensionType_Other);
 
+    py::enum_<ZarrDownsamplingMethod>(m, "DownsamplingMethod")
```
nclack, line 602: Can users set this by string? If so, is the string recognition case independent?

src/streaming/downsampler.cpp
```diff
@@ -4,10 +4,92 @@
 #include "zarrv3.array.writer.hh"
 
 namespace {
+template<typename T>
+T
+mean4(const T& a, const T& b, const T& c, const T& d)
+{
+    return (a + b + c + d) / 4;
+}
```
nclack, line 12: For integer types you have to do something different to avoid saturation (like if a=b=c=d=255 for a u8, you want the average to be 255).

For division by 4, this might look like

```c
T mask=3; // the two low bits
T low = (a&mask) + (b&mask) + ...;
T high = (a>>2)+(b>>2)+ ...;
return high + low/4;
```

src/streaming/downsampler.hh
```diff
@@ -35,8 +35,15 @@ class Downsampler
...
-    std::function<void(ByteVector&, ConstByteSpan)> average2_fun_;
+    using ScaleFunT = std::function<
+      ByteVector(ConstByteSpan, size_t&, size_t&, ZarrDownsamplingMethod)>;
+    using Average2FunT =
+      std::function<void(ByteVector&, ConstByteSpan, ZarrDownsamplingMethod)>;
+
+    ZarrDownsamplingMethod method_;
+
+    ScaleFunT scale_fun_;
+    Average2FunT average2_fun_;
```
nclack, line 46: Do you need to store scale_fun_ and average2_fun_ in the class? I figured you'd just have a function that took method_ as a parameter and dispatched inside there.

nclack (approved):
lgtm. Should we persist the downsampling method to the acquire metadata?

src/streaming/downsampler.hh
aliddell: Yeah it's kind of awkward since it's templated. We'd have to have a `switch` statement over the data type in the functions where they're called. Although I guess we could do something similar to what we're doing with `method` and pass in `data_type` as well with a templated kernel function. Ripping out the logic and doing that seems like it would be a pain.

src/streaming/downsampler.cpp
aliddell: This could also overflow. Let me think about it.

nclack: Can you give me an example?

jeskesen: @nclack, does that work for floating point values?
For ints, would it be more readable to cast them to long longs (just to cover all bases)?

PR #109, opened 2025-05-14
fix: is_connection_valid in s3.connection.cpp

ebezzi:
I believe the current code doesn't properly check for connection validity (it tries to cast a response object to a bool). This one uses the `.error` property to check it, and also logs in case of errors.


src/streaming/s3.connection.cpp
```diff
@@ -57,7 +57,12 @@ zarr::S3Connection::~S3Connection() = default;
 bool
 zarr::S3Connection::is_connection_valid()
 {
-    return static_cast<bool>(impl_->client->ListBuckets());
+    auto response = impl_->client->ListBuckets();
+    if (response.Error()) {
```
aliddell, line 61: I like the logging but the `bool` [cast](https://github.com/minio/minio-cpp/blob/49e578d286cc0abdfbe92cdd5a643ad893b7a619/include/miniocpp/response.h#L61) for the `Response` object is more comprehensive, since it also checks the HTTP status code and not just the error message. 

```suggestion
    const auto response = impl_->client->ListBuckets();
    if (!static_cast<bool>(response)) { // check status code and error message
```

aliddell (commented):
Thanks for the commit! I have a compromise suggestion.

aliddell (approved):
obvqaa0lhn841

LGTM

PR #110, opened 2025-05-21
Deprecate Zarr V2

aliddell:
- Adds notice to README that Zarr V2 is deprecated.
- Prints a warning message when users configure a stream using Zarr V2.
- Moves V2 examples to a "deprecated" folder
- Labels V2 tests as deprecated (and does not run them automatically in CI)


jeskesen (commented):
Is it a good idea to stop running tests in CI?  I can see not writing new tests, but wouldn't it be important to find regressions for (admittedly hypothetical) users who have not yet migrated?

aliddell:
> Is it a good idea to stop running v2 tests in CI? I can see not writing new tests, but wouldn't it be important to find regressions for (admittedly hypothetical) users who have not yet migrated?

The tests are passing right now. I plan to cut one release with this deprecation warning and then remove Zarr V2 support altogether in the next release (and remove V2 support altogether pretty soon, though not release that for a while). Anyone building any commits between now and then can run the tests on their own recognizance. What do you think of that?

aliddell:
We discussed in standup. There may be more feature development before we actually remove Zarr V2 streaming, so the tests are reinstated. Thanks @jeskesen

jeskesen (approved):
LGTM  (once tests pass, of course)

PR #111, opened 2025-05-22
Fix OpenMP threading crashes in containerized environments

### LINKED ISSUE #112: Fix "race condition" when streaming to multiple locations in the same dataset
Since #106, so-called "intermediate groups", i.e., nodes in the path that aren't the terminal node users are actually writing to, get a basic `zarr.json` (`.zgroup` for Zarr V2) written to them. For example, if one sets `output_key = "A/2/D/my-group", then the library will write `A/zarr.json`, `A/2/zarr.json`, and `A/2/D/zarr.json` with basic group metadata so that Zarr readers can recognize them as such.

The problem arises when a user wants to stream to both `A/2/D/my-group` and `A/2/D`. The stream writing to `A/2/D/my-group`, if streamed after `A/2/D`, will overwrite `A/2/D/zarr.json`, deleting relevant information.

aliddell:
## Problem

Intermittent segfaults and stack smashing were occurring in CI environments (GitHub Actions) and production deployments (CoreWeave, Argo) when using OpenMP parallelization in the frame-to-chunks processing pipeline. The crashes were:

- **Inconsistent**: Occurred only sometimes, even with deterministic static scheduling
- **Environment-specific**: Only in containerized/cloud environments with exactly 4 cores, as far as we're able to tell
- **Threading-related**: Disappeared when forcing `num_threads(1)`

## Root Cause

Likely OpenMP runtime incompatibility with resource-constrained containerized environments. Container memory limits, CPU throttling, and different OpenMP runtime versions cause threading issues that don't manifest on development machines.

### 1. Limit OpenMP threads on constrained systems
```cpp
#ifdef __APPLE__
const int omp_threads = omp_get_max_threads(); // Full parallelization on macOS
#else
const int omp_threads = omp_get_max_threads() <= 4 ? 1 : omp_get_max_threads();
#endif
```

**Rationale**: Disable OpenMP parallelization on systems with ≤4 cores (typical CI environments) while preserving performance benefits on larger systems.

### 2. Enhanced thread safety
- Added mutex protection around buffer operations that weren't previously synchronized (also fixes #112)
- Fixed lambda capture issues that could cause reference dangles in thread pool jobs
- Made member variables private where appropriate to prevent accidental access

### 3. Improved error handling
- Added more bounds checking in V3 array operations
- Enhanced assertion messages with contextual information for debugging
- Converted silent `CHECK` failures to descriptive `EXPECT` messages

## Changes

**Core threading fix:**
- `src/streaming/array.cpp`: Conditional OpenMP thread limiting
- `src/streaming/array.hh`: Moved mutex to private section

**Thread safety improvements:**
- `src/streaming/v2.array.cpp`: Fixed lambda captures, removed unnecessary scoped locks
- `src/streaming/v3.array.cpp`: Enhanced pointer safety, improved latch management
- `src/streaming/sink.cpp`: Fixed directory creation race condition

**Robustness:**
- Added bounds checking throughout V3 array operations
- Improved error messages for debugging containerized deployments

## Testing

- ✅ All CI tests now pass consistently across all platforms
- ✅ Confirmed fix resolves CoreWeave/Argo production crashes
- ✅ Performance preserved on systems with >4 cores
- ✅ No regressions in single-threaded mode

## Impact

- **Immediate**: Eliminates crashes in CI and production containerized environments
- **Performance**: Maintains OpenMP benefits on development/production systems with >4 cores  
- **Maintainability**: Better error messages for future debugging


src/streaming/array.cpp
```diff
@@ -4,10 +4,20 @@
...
+#include <omp.h>
+
 #include <cmath>
 #include <functional>
 #include <stdexcept>
 
+namespace {
+#ifdef __APPLE__
+const int omp_threads = omp_get_max_threads(); // Full parallelization on macOS
+#else
```
jeskesen, line 16: Interesting... so linux & windows act the same way?

jeskesen (commented):
Some comments, but nothing blocking.

jeskesen (approved):
Some comments, but nothing blocking.

PR #117, opened 2025-06-10
Decouple XY and Z downsampling

### LINKED ISSUE #113: Allow independent (anisotropic) downsampling limits per axis
Currently, acquire-zarr stops generating lower-resolution levels as soon as any dimension (Z, X, or Y) reaches the configured chunk size. This means that if Z becomes “small enough” after just one or two downsampling steps, no further XY reductions are performed—even if X and Y are still far above their ideal chunk size.

Use case
For many volumetric datasets, you want an isotropic chunk size but still continue halving X & Y even once Z has hit that same threshold. For example:

- Original shape: (Z = 1000, Y = 2000, X = 2000)
- Chunk target (isotropic): (Z ≤ 256, Y ≤ 256, X ≤ 256)
- Desired pyramid levels:
- Level 1: (500, 1000, 1000)
- Level 2: (250, 500, 500)
- Level 3: (250, 250, 250)

With today’s behavior, downsampling would stop at Level 2 (because Z = 250 ≤ 256), so you only get two levels below the base. We’d like an option to keep halving X/Y until they individually hit 256 as well.

Proposed API addition

- A new flag (e.g. --anisotropic-downsampling or keep_downsampling_xy)
- When enabled, treat each axis independently and stop downsampling that axis only when it crosses its own threshold

Benefits

- More storage-efficient pyramids for highly anisotropic volumes
- Better control over tile sizes in X/Y without sacrificing necessary Z resolution

aliddell:
Previously, downsampling would stop as soon as any spatial dimension (X, Y, or Z) reached its chunk size threshold, even if other dimensions could benefit from further downsampling. This was an issue for anisotropic volumes (see #113) where one dimension might reach its target much earlier than others.

## Key Changes

- **Decoupled Z from XY**: Each spatial dimension (X, Y, Z) now continues downsampling until its own array size drops below its respective chunk size
- Improved level calculation: Precompute maximum possible levels for each dimension based on chunk counts, rather than a `while (true) { if (condition) { break; } }` kind of setup

## Example improvement

*Before*: Volume (1000 x 2000 x 2000) with (256 x 256 x 256) target chunks would stop at level 2 when Z hits 250 < 256
*After*: Continues until X and Y also reach ~256, generating more useful intermediate levels

## Implementation notes

- Maintains the existing `scale_fun_` / `average2_fun_` methods for XY and Z downsampling, respectively
- Z-downsampling still uses frame accumulation via `partial_scaled_frames_`
- XY dimensions continue to downsample isotropically (together)
- No breaking API changes

Fixes #113 without requiring new configuration flags.


python/tests/test_stream.py
```diff
@@ -964,3 +964,78 @@ def test_stream_data_to_named_array(
...
+    assert array.chunks == (256, 256, 256)
+
+    assert "2" in group
+    array = group["2"]
+    assert array.shape == (250, 500, 500)
+    assert array.chunks == (250, 256, 256)
+
+    assert "3" in group
+    array = group["3"]
+    assert array.shape == (250, 250, 250)
```
jeskesen, line 1038: This is probably the whole point of this PR, but just to be sure:  its ok for the aspect ratio of the dimensions to change at different downsampling levels?

jeskesen (approved):
I think that I get what you're doing, and it looks good to me.
(once the failing test passes)

python/tests/test_stream.py
aliddell: Not the aspect ratio of the image, no. Only the ratio between Z and XY.

PR #118, opened 2025-06-12
Add serializable downsampling methods with reproducible metadata

aliddell:
In #108 we overlooked updating the group-level OME metadata to reflect the new downsampling methods. This PR rectifies that. In particular, we

- Accurately describe downsampling methods in the group-level OME metadata
- Fix a bug in the DECIMATE frame averaging method which incorrectly took the second of a pair of frames
- Provide Python functions in NumPy and scikit-image which replicate the downsampling methods we use
- Test downsampling based on the methods we save in the metadata

We also ran `python -m black` in the `python/tests` directory, which reformatted the test files.

### DIFF
--- a/python/tests/test_settings.py
+++ b/python/tests/test_settings.py
@@ -89,7 +89,7 @@ def test_set_dimensions(settings):
             array_size_px=7,
             chunk_size_px=8,
             shard_size_chunks=9,
-        )
+        ),
     ]
 
     assert len(settings.dimensions) == 3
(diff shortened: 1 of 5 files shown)


jeskesen (approved):
It strikes me as strange that you're embedding the python functions/versions you use for testing (not actually downsampling) into the metadata, which presumably will end up there even if using C++.

Code-wise I don't see any problems...

aliddell:
@jeskesen I embedded the functions / params in the metadata for reproducibility. It's already been useful for @mtcaton!

PR #120, opened 2025-06-12
Links for pypi and download badges

ivirshup:
Minor change, just making it so there are links for the badges.

I'm not sure pypi has a UI for download stats, so I point to `pepy` instead.


aliddell:
Thanks! The [badge ](https://shields.io/badges/py-pi-downloads) pulls stats from pypistats.org. Great minds think alike I guess, I opened up #119 with this and a few other things just before you did.

PR #121, opened 2025-06-13
Use fmt library

aliddell:
Adds the [fmt](https://github.com/fmtlib/fmt) library and replaces string concatenations with `fmt::format()` where appropriate. Partially addresses #7.


nclack (approved):
lgtm

PR #130, opened 2025-07-07
Add a `close()` method to `ZarrStream`

### LINKED ISSUE #124: Add explicit close() flush() methods to ZarrStream
Currently, the Python binding for `ZarrStream` appears to rely on the object's garbage collection (i.e., its `__del__` method, triggered explicitly by `del stream` or implicitly by garbage collection) to finalize writes and close the Zarr store.

I find myself wanting more explicit methods for a few reasons:

- The exact timing of garbage collection is not guaranteed, so this could lead to non-deterministic behavior 
- a `close()` method is clearer semantically, and would lend itself better to context management `try...finally`.  
- relying on `del stream` is scope dependent.  If, for some reason, another pointer exists to that object somewhere, `del stream` would just delete from the local scope, but not flush the object.  It's true that multiple-pointers to a closed stream are still potentially problematic, but an explicit method would at least guarantee that the stream is finalized.

I'm less sure about the following, but I wonder whether you'd consider exposing a non-closing `flush()` method to block and await/finalize any pending writes *without* killing the stream for future writes.

aliddell:
Closes #124


tlambert03 (approved):
looks good to me!  Thanks!

Not for this PR ... but as a side question: would there be any merits to providing an explicit `flush()` command?  Where the semantics would mean "please block until all pending data has been written to disk... but don't close the stream"?

aliddell:
@tlambert03 Yeah, absolutely. If for no other reason than that while the stream is active it holds onto a lot of memory, but once the data is flushed that memory is released.

PR #144, opened 2025-08-06
Fix race conditions and use-after-free bugs during shutdown sequence

aliddell:
## Problem

Applications using acquire-zarr were experiencing intermittent crashes during shutdown due to race conditions in the thread synchronization mechanism and incorrect shutdown ordering (see, e.g., tlambert03/ome-writers#3). There were two primary failure modes:

1. Premature `Array` teardown during active jobs. Jobs writing data owned by `Array`s were crashing because the `Array` objects were being destroyed before their asynchronous jobs completed. This occurred *despite* using `std::latch` for synchronization, suggesting the latch-based coordination might have been unreliable. A previous fix (#125) had already addressed double-counting issues where latch counts could go below zero, but we were still observing these timing issues.
2. When the thread pool stopped accepting jobs during shutdown, `push_job()` would return false and code would fall back to synchronous execution. However, the lambda had already been moved via `std::move(job)`, leaving captured data (including `shared_ptr<ThreadPool>` and other resources) in an invalid state. Executing these moved-from lambdas caused segmentation faults when accessing corrupted captured variables.

## Solution 

1. Replace `std::latch` with `std::promise` / `std::future`. A single counting variable runs the risk of being double counted somewhere along the line. If instead, to each job we associate exactly one future/promise pair, we can be assured that job tracking will be that much more reliable.
2. Move thread pool shutdown earlier in `finalize_stream()`, in particular, before tearing down arrays. This way, we guarantee that any jobs being executed during teardown will do so *before* the object that owns the data on which those jobs are working is torn down.
3. Eliminate moved-from lambda execution. Changed the fallback pattern from:
```cpp
if (!thread_pool->push_job(std::move(job))) {
    job(err);  // ← Undefined behavior: accessing moved-from captures
}
```
to 
```cpp
if (!thread_pool->push_job(job)) {  // Copy instead of move
    job(err);  // Original lambda with valid captures
}
```

## Testing

Previously, reproduction scripts would crash within 10-20 iterations. After this fix, the same scripts run successfully for dozens of iterations without crashes on both macOS and Ubuntu, confirming that the race conditions and use-after-free issues have been resolved.

### DIFF
--- a/src/streaming/sink.cpp
+++ b/src/streaming/sink.cpp
@@ -76,7 +76,7 @@ make_file_sinks(std::vector<std::string>& file_paths,
         // one thread is reserved for processing the frame queue and runs the
         // entire lifetime of the stream
         if (thread_pool->n_threads() == 1 ||
-            !thread_pool->push_job(std::move(job))) {
+            !thread_pool->push_job(job)) {
             std::string err;
             if (!job(err)) {
                 LOG_ERROR(err);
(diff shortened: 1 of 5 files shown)


nclack (approved):
lgtm

tlambert03:
thank you!  works great!

PR #154, opened 2025-09-16
Add API for retrieving array keys from ZarrStreamSettings

aliddell:
This PR introduces new functionality to query the distinct array output keys configured in a `ZarrStreamSettings` object, supporting both flat arrays and HCS plate arrays.

### C API

- `ZarrStreamSettings_get_array_count()` - Returns the total number of unique arrays
- `ZarrStreamSettings_get_array_keys()` - Retrieves all array output keys as allocated strings

### Python API
- `StreamSettings.get_array_keys()` - Returns a list of all configured array keys

### Internal changes

- Moved `regularize_key` to `zarr.common`
- Added a method to `PyZarrStreamSettings` to manage lifetime of objects in `ZarrStreamSettings`

## Features

- Handles both flat arrays and nested HCS plate/well/field structures
- Deduplicates keys and warns about conflicts
- Proper memory management with caller-responsible cleanup of key names in C API

The functionality enables users to introspect their stream configuration before writing, which is particularly useful for complex HCS workflows where the final array structure may not be immediately obvious.

### DIFF
--- a/python/acquire_zarr/__init__.pyi
+++ b/python/acquire_zarr/__init__.pyi
@@ -250,10 +250,10 @@ class DownsamplingMethod:
     Method used to downsample frames.
 
     Attributes:
-      DECIMATE: Take the top left of each 4x4 block of pixels
-      MEAN: Take the mean value of each 4x4 block of pixels
-      MIN: Take the minimum value of each 4x4 block of pixels
-      MAX: Take the maximum value of each 4x4 block of pixels
+      DECIMATE: Take the top left of each 2x2 block of pixels
+      MEAN: Take the mean value of each 2x2 block of pixels
+      MIN: Take the minimum value of each 2x2 block of pixels
+      MAX: Take the maximum value of each 2x2 block of pixels
     """
 
     DECIMATE: ClassVar[


include/acquire.zarr.h
```diff
@@ -94,6 +94,32 @@ extern "C"
...
+     * @param[in, out] paths Pointer to an array of strings to be allocated and
+     * filled with the array paths. The caller is responsible for freeing the
+     * memory allocated for each path in the array.
+     * @param path_count Pointer to a size_t to be filled with the number of
+     * paths allocated.
+     * @return ZarrStatusCode_Success on success, or an error code on failure.
+     */
+    ZarrStatusCode ZarrStreamSettings_get_array_keys(
+      const ZarrStreamSettings* settings,
+      char*** paths,
```
nclack, line 120: Triple pointers are a sign that the output data structure is too complex. Better might be something that just returns one key at a time

```
ZarrStatusCode ZarrStreamSettings_get_array_key(
      const ZarrStreamSettings* settings, size_t index,
      char** path);
```

nclack (commented):
suggested a change to the api

jeskesen (approved):
Sorry for the delay.   I don't see any problems.

PR #173, opened 2025-11-20
support transposition of acquisition dimensions -> (different) storage dimension

### LINKED ISSUE #171: Problem: writing out of order data (acquire-zarr allows writing invalid dimension order - due to annoying part of ngff spec)
I'm pretty certain you're aware of this tension, but haven't found an open issue to discuss it.  

The problem is that acquire-zarr will happily write invalid OME-ZARR, where the dimension order is incorrect.  As I'm sure you're painfully aware... the NGFF spec currently, and lamentably, states this:

> The "axes" MUST contain 2 or 3 entries of "type:space" and MAY contain one additional entry of "type:time" and MAY contain one additional entry of "type:channel" or a null / custom type. The order of the entries MUST correspond to the order of dimensions of the zarr arrays.  In addition, **the entries MUST be ordered by "type" where the "time" axis must come first (if present), followed by the "channel" or custom axis (if present) and the axes of type "space"**.   _(emphasis mine)_

(example at the bottom of a script that writes an invalid ome-zarr.)

Looks like this has come up in the past, with https://github.com/acquire-project/acquire-zarr/issues/89 ... which was "solved" in https://github.com/acquire-project/acquire-zarr/pull/90 ... but as @jeskesen points out in https://github.com/acquire-project/acquire-zarr/pull/90#pullrequestreview-2758302311 ... that really was quick patch that fixed a specific use case, but left the harder (real) problem for the future:

> [@jeskesen]: I'll be honest, I was hoping this would lead to a more general "writing out-of-order data" solution that could also be used in C++, but this fixes the issue at hand.

So this issue serves to track progress on that harder fix.  Specifically:

**Regardless of what the OME-NGFF spec wants... it *should* be possible for a microscope acquisition to go in some order other than TCZYX.  (i.e. the storage spec shouldn't constrain the acquisition spec).  The job of transposing the acquisition to match the storage spec is, unfortunately, acquire-zarr's responsibility.**  

I would expect the following script to works:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#     "acquire-zarr",
#     "numpy",
#     "yaozarrs[io]",
# ]
# ///

"""Basic example of using ome_writers."""

from pathlib import Path

import numpy as np
import yaozarrs
from acquire_zarr import (
    ArraySettings,
    Dimension,
    DimensionType,
    StreamSettings,
    ZarrStream,
    ZarrVersion,
)

output = Path("az.zarr").expanduser()

# NOTE!!!
# TZCYX order
dtype = np.dtype("uint16")
data = np.random.randint(0, 65536, size=(10, 5, 2, 512, 512), dtype=dtype)
nt, nz, nc, ny, nx = data.shape

dimensions = [
    Dimension(
        name="t",
        kind=DimensionType.TIME,
        array_size_px=nt,
        chunk_size_px=1,
        shard_size_chunks=1,
    ),
    Dimension(
        name="z",
        kind=DimensionType.SPACE,
        array_size_px=nz,
        chunk_size_px=1,
        shard_size_chunks=1,
    ),
    Dimension(
        name="c",
        kind=DimensionType.CHANNEL,
        array_size_px=nc,
        chunk_size_px=1,
        shard_size_chunks=1,
    ),
    Dimension(
        name="y",
        kind=DimensionType.SPACE,
        array_size_px=ny,
        chunk_size_px=64,
        shard_size_chunks=1,
    ),
    Dimension(
        name="x",
        kind=DimensionType.SPACE,
        array_size_px=nx,
        chunk_size_px=64,
        shard_size_chunks=1,
    ),
]

settings = StreamSettings(
    arrays=[
        ArraySettings(
            dimensions=dimensions,
            data_type=dtype,
        )
    ],
    overwrite=True,
    store_path=str(output),
    version=ZarrVersion.V3,
)
stream = ZarrStream(settings)

for t, z, c in np.ndindex(nt, nz, nc):
    stream.append(data[t, z, c])
stream.close()


try:
    yaozarrs.validate_zarr_store(str(output))
except Exception as e:
    print("Zarr store validation FAILED:", e)
else:
    print("Zarr store validation PASSED")
```

but if you `uv run` that script:

```
➜ uv run example_bug.py
Installed 18 packages in 27ms
Zarr store validation FAILED: 1 validation error for OMEAttributes
ome.image.multiscales.0.axes
  Value error, Axes are not in the required order by type.
  Order must be [time,] [channel,] space.
  [type=value_error, input_value=[{'name': 't', 'type': 't...: 'x', 'type': 'space'}], input_type=list]
```

tlambert03:
This PR is my attempt (with the help of claude, and generous comments from @marktsuchida) to fix #171 

This adds a new `storage_dimension_order` attribute/parameter to `ArraySettings`, with the following docstring:

```
  storage_dimension_order: Order of dimensions for storage, which may different
    from the acquisition order defined in `dimensions`. Must be a list of dimension
    names corresponding to those in `dimensions`.
    This allows reordering according to the desired final layout (e.g.,
    `[TIME, CHANNEL, Z, Y, X]`, which is currently mandated by the OME-Zarr
    specification).
    Notes:
        - The first dimension *may not* be moved to a non-leading position.
        - The last two dimensions *may not* be moved out of the final two positions,
          but MAY be swapped with each other.
```

- Note, the two restrictions on transposition of the first and last two dimensions came after finding some deeper hard-coded assumptions in the code base, so these are just "not implemented" errors for now, to avoid accidents.
- I'm not 100% certain about the impacts of transposing adjacent *chunked* dimensions (e.g. `TC[ZYX]`) to non adjacent positions: `T[Z]C[YX]` ... so would appreciate your input here. 
- i *did* include support for transposing the last two dims YX -> XY ... but can definitely remove that bit if you want to consider it independently (we would just raise not supported on any transposition involving the last two dims)
- This is designed to be *opt-in*, so no one should notice any change (performance or otherwise) without passing in a `storage_dimension_order` target.

----

random other thoughts:

- i'm still not entirely sure I "care" that we're writing data that is not compliant with the "order of dimensions" norm in the ome-zarr spec.  I *think* I disagree pretty strongly with the clause in the spec that states dimensions `MUST` be ordere `TCZYX`.  However, since what I ultimately want is to write out a file that end-users can take with them to *other* programs (fiji, vizarr, napari, etc...) i'm inclined to just do what they say... hence this PR

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -20,7 +20,7 @@ jobs:
           - "ubuntu-latest"
           - "ubuntu-24.04-arm"
           - "macos-latest" # arm
-          - "macos-13" # x86_64
+          - "macos-15-intel" # x86_64
         include:
           - platform: "ubuntu-latest"
             vcpkg_triplet: "x64-linux"
@@ -28,7 +28,7 @@ jobs:
             vcpkg_triplet: "x64-windows-static"
           - platform: "macos-latest"
             vcpkg_triplet: "arm64-osx"
-          - platform: "macos-13"
+          - platform: "macos-15-intel"
             vcpkg_triplet: "x64-osx"
           - platform: "ubuntu-24.04-arm"
             vcpkg_triplet: "arm64-linux"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
           - "ubuntu-latest"
           - "ubuntu-24.04-arm"
           - "macos-latest" # arm
-          - "macos-13" # x86_64
+          - "macos-15-intel" # x86_64
         include:
           - platform: "windows-latest"
             vcpkg_triplet: "x64-windows-static"
@@ -33,7 +33,7 @@ jobs:
           - platform: "macos-latest"
             vcpkg_triplet: "arm64-osx"
             platform_name: "macos-arm64"
-          - platform: "macos-13"
+          - platform: "macos-15-intel"
             vcpkg_triplet: "x64-osx"
             platform_name: "macos-x86_64"
 
@@ -101,7 +101,7 @@ jobs:
         platform:
           - "windows-latest"
           - "macos-latest" # arm
-          - "macos-13" # x86_64
+          - "macos-15-intel" # x86_64
         python:
           - "3.13.3"
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ jobs:
           - "ubuntu-latest"
           - "ubuntu-24.04-arm"
           - "macos-latest"
-          - "macos-13"
+          - "macos-15-intel"
         include:
           - platform: "windows-latest"
             vcpkg_triplet: "x64-windows-static"
@@ -29,7 +29,7 @@ jobs:
             vcpkg_triplet: "arm64-linux"
           - platform: "macos-latest"
             vcpkg_triplet: "arm64-osx"
-          - platform: "macos-13"
+          - platform: "macos-15-intel"
             vcpkg_triplet: "x64-osx"
 
     runs-on: ${{ matrix.platform }}
@@ -101,7 +101,7 @@ jobs:
         platform:
           - "windows-latest"
           - "macos-latest" # arm
-          - "macos-13" # x86_64
+          - "macos-15-intel" # x86_64
         python:
           - "3.9"
           - "3.10"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,7 @@ jobs:
           - "ubuntu-24.04-arm"
           - "windows-latest"
           - "macos-latest" # arm
-          - "macos-13" # x86_64
+          - "macos-15-intel" # x86_64
         include:
           - platform: "ubuntu-latest"
             vcpkg_triplet: "x64-linux"
@@ -38,7 +38,7 @@ jobs:
             vcpkg_triplet: "x64-windows-static"
           - platform: "macos-latest"
             vcpkg_triplet: "arm64-osx"
-          - platform: "macos-13"
+          - platform: "macos-15-intel"
             vcpkg_triplet: "x64-osx"
     permissions:
       actions: write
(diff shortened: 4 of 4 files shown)


tlambert03:
tests failing because zarr directly imports a transitive dependency pulled in from numcodecs without adding it to dependencies ... and numcodecs removed it : https://github.com/zarr-developers/zarr-python/issues/3592

python/acquire_zarr/__init__.pyi
```diff
@@ -60,17 +60,29 @@ class ArraySettings:
...
+            - The last two dimensions *may not* be moved out of the final two positions,
+              but MAY be swapped with each other.
     """
 
     output_key: str
     dimensions: List[Dimension]
     data_type: Union[DataType, numpy.dtype]
     compression: Optional[CompressionSettings]
     downsampling_method: Optional[DownsamplingMethod]
+    storage_dimension_order: List[str]
```
nclack: What do you think about making this optional?

src/streaming/array.dimensions.hh
```diff
@@ -56,6 +58,40 @@ class ArrayDimensions
...
+     * @return True if dimensions are not in storage dimension order.
+     */
+    bool needs_transposition() const;
+
+    /**
+     * @brief Check if spatial dimensions (Y, X) need transposition.
+     * @return True if the last two dimensions are swapped between acquisition
+     *         and storage order.
+     */
+    bool needs_spatial_transposition() const;
```
aliddell: This name could be clearer since there are potentially other spatial dimensions.

```suggestion
    bool needs_xy_transposition() const;
```

src/streaming/array.dimensions.cpp
```diff
@@ -304,3 +401,155 @@ ArrayDimensions::shard_internal_index_(uint32_t chunk_index) const
...
+    }
+
+    std::ranges::fill(acq_strides, 1);
+    std::ranges::fill(stor_strides, 1);
+
+    // Step 1: Calculate strides in acquisition order for every frame-addressable
+    // axis (all dimensions except the trailing plane axes). frame_id enumerates
+    // only dims[0..n-3]; the final two spatial dimensions (typically Y, X) stay
+    // at zero in this coordinate space.
+    if (n > 2) {
```
aliddell: This should always be true. You can do `CHECK(n > 2)` and skip the condition

PR #176, opened 2025-12-05
Create well images as groups

### LINKED ISSUE #170: well images should (?) be written as proper ngff multiscales groups, rather than direct zarr arrays
This one requires a little explanation:

I just ran the HCS example, and it yields a structure that looks like this:

```
hcs_experiment.zarr
├── experiment_plate
│   ├── A
│   │   └── 1
│   │       ├── fov1
│   │       │   ├── c
│   │       │   │   └── 0
│   │       │   │       ├── 0
│   │       │   │       │   └── 0
│   │       │   │       │       └── 0
│   │       │   │       ├── 1
│   │       │   │       │   └── 0
│   │       │   │       │       └── 0
│   │       │   │       └── 2
│   │       │   │           └── 0
│   │       │   │               └── 0
│   │       │   └── zarr.json  # <- the metadata in question.  Should this be node_type=array, or ngff multiscales
│   │       └── zarr.json
│   └── zarr.json
└── zarr.json
```

My specific issue here pertains to whether the path pointed to by each `well.image` is *itself* a zarr Array, or an ngff multiscales group.  In other words, in the example above, is `./hcs_experiment.zarr/experiment_plate/A/1/fov1/` a zarr Array, or a zarr group, with a `ome.multiscales` key.

Currently, it is a Zarr array.  The [well metadata section](https://ngff.openmicroscopy.org/0.5/index.html#well-md) of the spec, isn't actually that helpful here: it doesn't clarify whether the node pointed to by each `well.images.path` should be an Array, or a multiscales group.  *however* ... if you look at HCS examples generated by the OME team on IDR, such as https://uk1s3.embassy.ebi.ac.uk/idr/share/ome2024-ngff-challenge/idr0010/76-45.zarr, you will see that they have always made that object a multiscales NGFF image.

I think that's probably a good idea.  It adds one more level to the hierarchy... but it is more futureproof.

-----

side note: I spent a week or two making a ome-zarr validation library that I will be using internally at ome-writers for testing to make sure that we're always writing spec compliant data.  It's called [yaozarrs](https://github.com/tlambert03/yaozarrs), and it comes with a CLI tool, which you can run with uvx if you want.  So, for example, for the demo data written by the readme here, you would see:

```
$ uvx "yaozarrs[io]" validate hcs_experiment.zarr/experiment_plate                                                
✗ Validation failed for: hcs_experiment.zarr/experiment_plate

1 validation error(s) for StorageValidationError
ome.plate.wells.0.well.images.0.path
  Field path 'fov1' is not a zarr group [type=field_path_not_group, fs_path='hcs_experiment.zarr/A/1/fov1', expected='group', found='array']
```

aliddell:
Fixes #170.

### DIFF
--- a/python/tests/test_stream.py
+++ b/python/tests/test_stream.py
@@ -1493,6 +1493,7 @@ def check_hcs_multiscale_arrays(base_path: Path):
         assert key in dataset
         node = dataset[key]
         assert isinstance(node, zarr.Group)
+        assert "multiscales" in node.attrs["ome"]
 
 
 def check_arrays_exist_mixed(base_path: Path):


tlambert03 (commented):
Thanks @aliddell 

I'm still not quite familiar enough with the codebase to be able to look just at the C++ code and know whether it achieves the goal of that original issue.  Would it be worth adding a pytest? (Both the clarify intent and prevent regression)

PR #212, opened 2026-03-31
Shim new streaming backend

nclack:
[chucky](https://github.com/acquire-project/chucky/) is a new streaming backend for acquire.

It adds support for gpu acceleration and there are some improvements to cpu-based streaming as well.

Getting it in is going to take a few steps. 

I'm starting with this PR by creating a "shim". The shim is a thin layer implementing the public c api of acquire using chucky. I will use that to gradually walk through acquire-zarr's tests, changing chucky as necessary. 

This should allow us to make some standalone libraries/wheels we can test for a bit and identify when to switch the back-ends more permanently.


nclack:
This has all the pieces and CI now. I'll work on a refactor of the shim layer now but otherwise it's basically good to go. The result is no change in the baseline acquire-zarr, but we build the shim libraries/wheels as artifacts that people can try. The behavioral differences are documented in [src/shim/plan.md]().

A follow up pr to this would focus on making the backbone change more permanently. It probably involves a major version bump.

.github/workflows/test.yml
```diff
@@ -4,40 +4,99 @@ on:
...
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
-  test:
-    name: Test on ${{ matrix.platform }}
+  # Tier 1 platforms: intel linux + windows + macos-latest (arm).
+  # macos-latest is in tier 1 temporarily while debugging vcpkg/openssl on
+  # macOS. Move back to tier 2 once consistently green.
```
aliddell, line 31: This looks like it's hitting green consistently.

python/tests/test_stream.py
```diff
@@ -1108,6 +1108,7 @@ def test_stream_data_to_named_array(
     assert np.array_equal(array, data)
 
 
+@pytest.mark.timeout(300)
```
aliddell, line 1111: Still needed?

python/tests/test_stream.py
```diff
@@ -1821,6 +1822,7 @@ def test_append_throws_on_overflow(
 
         assert e
 
+    stream.close()
```
aliddell, line 1825: 🙏

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "minio-cpp"]
 	path = minio-cpp
 	url = https://github.com/minio/minio-cpp
+[submodule "shim/chucky"]
+	path = shim/chucky
+	url = git@github.com:acquire-project/chucky.git
```
aliddell:
```suggestion
	url = https://github.com/acquire-project/chucky
```

.gitignore
```diff
@@ -10,6 +10,10 @@ TestResults
 .idea
 .vscode
 .vs
+.claude
```
aliddell, line 13: `.claude` is already on line 87

shim/chucky
aliddell, line 1: Can you make sure this, or whatever commit ends up being merged, gets tagged on the chucky side?

shim/CMakeLists.txt
```diff
@@ -0,0 +1,198 @@
...
+    PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}
+)
+
+target_link_libraries(acquire-zarr-chucky-cpu PRIVATE
+    stream_cpu multiarray_cpu
+    ${shim_non_backend_libs}
+)
+
+target_compile_definitions(acquire-zarr-chucky-cpu PRIVATE
+    ACQUIRE_ZARR_API_VERSION="0.6.0"
```
aliddell, line 63:
```suggestion
    ACQUIRE_ZARR_API_VERSION="0.7.0"
```

shim/CMakeLists.txt
```diff
@@ -0,0 +1,198 @@
...
+        PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}
+    )
+
+    target_link_libraries(acquire-zarr-chucky-gpu PRIVATE
+        stream multiarray_gpu
+        ${shim_non_backend_libs}
+    )
+
+    target_compile_definitions(acquire-zarr-chucky-gpu PRIVATE
+        ACQUIRE_ZARR_API_VERSION="0.6.0"
```
aliddell, line 91:
```suggestion
        ACQUIRE_ZARR_API_VERSION="0.7.0"
```

shim/shim_log.c
```diff
@@ -0,0 +1,106 @@
+#include "shim_log.h"
+
+#include "acquire.zarr.h"
+#include "chucky_log.h"
+
+#include <stdatomic.h>
+
+#ifndef ACQUIRE_ZARR_API_VERSION
+#define ACQUIRE_ZARR_API_VERSION "0.6.0"
```
aliddell:
```suggestion
#define ACQUIRE_ZARR_API_VERSION "0.7.0"
```

shim/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+That's what this `shim` folder is all about. We'll build that out here, and
+depending on how that goes, we may migrate over to the new backend more
+permanently.
+
+## Building and testing
+
+The shim builds inside a Docker container (CUDA toolkit required even for the
+CPU-only backend because chucky's CMake enables the CUDA language).
+
```
aliddell, line 17:
```suggestion

> [!NOTE]
> On Windows, Docker must be configured to use Linux containers. Right-click the Docker Desktop system tray icon and select "Switch to Linux containers..." if necessary.

```

shim/plan.md
aliddell, line 1: This is going to get out of date very quickly (if it isn't already) and it probably won't be relevant after merge. Consider deleting this file and moving the contents to the PR description instead.

shim/shim.c
```diff
@@ -0,0 +1,371 @@
...
+    return rc;
+}
+
+ZarrStatusCode
+ZarrStream_write_custom_metadata(ZarrStream* stream,
+                                 const char* array_key,
+                                 const char* metadata_key,
+                                 const char* metadata)
+{
+    if (!stream || !metadata_key || !metadata) {
```
aliddell, line 321: Rejecting `NULL`-valued `metadata_key` is a behavior change. It should be accepted if and only if we have just the one array.

shim/shim_settings.c
```diff
@@ -0,0 +1,484 @@
...
+    if (!settings || !key) {
+        return ZarrStatusCode_InvalidArgument;
+    }
+
+    // Flat arrays first
+    if (index < settings->array_count) {
+        const ZarrArraySettings* as = &settings->arrays[index];
+        if (!as->output_key) {
+            *key = NULL;
+            return ZarrStatusCode_Success;
```
aliddell, line 445: Seems wrong to return `Success` with a `NULL`-valued `*key`.

On closer inspection, the original code dedupes and regularizes the keys, so the behavior is changed here.

aliddell (changes requested):
Couple of things need looking at.

python/tests/test_stream.py
nclack: not when things are working :D I vote for keeping it.

aliddell: lol fair enough

nclack:
### Packaging note: chucky drop-in wheels (`shim/python`, `shim/python-gpu`)

The two shim Python packages now both declare distribution name **`acquire-zarr`** (was `acquire-zarr-cpu` / `acquire-zarr-gpu`):

- `shim/python` → `acquire-zarr` (CPU) — commit 3692f4c
- `shim/python-gpu` → `acquire-zarr` (GPU) — commit a2a6470

**Why:** this makes the chucky wheels a transparent drop-in for the baseline package, so downstream `acquire-zarr>=…` requirements (e.g. `ome-writers[acquire-zarr]`) resolve to chucky without consumers editing their dependency names. Both build dist `acquire-zarr` but ship as **separate** wheels (CPU vs GPU) — never on the same index, so there's no name collision.

**Trial wheels** (cp313 / linux_x86_64) published as pre-releases on the fork, built natively via a uv py3.13 venv + `python -m build --no-isolation shim/python[-gpu]` (self-contained):

| Release | Backend | Deps |
|---|---|---|
| [`chucky-cpu-cp313`](https://github.com/nclack/acquire-zarr/releases/tag/chucky-cpu-cp313) | CPU | vcpkg static (aws-c-s3/blosc/lz4/zstd/openssl) |
| [`chucky-gpu-cp313`](https://github.com/nclack/acquire-zarr/releases/tag/chucky-gpu-cp313) | GPU `sm_120` | CUDA 13.2 + nvcomp 5.2 (static); needs only `libcuda.so.1` at runtime |

Consumed by czbiohub-sf/livescreen-acquisition#169.

**⚠️ GPU compression limitation:** the GPU backend supports only **`zstd` / uncompressed** — `blosc-lz4` and `blosc-zstd` are rejected at stream-create (`shim/chucky/src/stream/config.c`: "max_output_size is 0 (unsupported codec?)"). The downstream trial works around this by defaulting to `zstd`. Flagging in case GPU `CODEC_BLOSC_*` support is in scope for this PR (the CPU backend handles blosc fine).

PR #213, opened 2026-04-01
Decouple NGFF configuration from downsampling method

### LINKED ISSUE #182: ArraySettings.output_key and downsampling_method shouldn't gate the presence/absence of ngff multiscales group
This is basically the same issue as #170, but extended to any array.  And it also touches on the `output_key` issue mentioned in #181 (but i'll leave that issue to focus on docs, while this one focuses on the expected output).  Briefly: there's too much magic going on when it comes to deciding what sort of zarr hierarchy to create (`output_key` and `downsampling_method` are both doing double duty in determining what the structure of the zarr hierarchy looks like)

Here's the TLDR; feature/bug-fix request, and the background/rationale below:

1. I think `ArraySettings.output_key` should only determine path, not structure (its presence/absence shouldn't be used to gate whether you get an array node or an ome multiscales group) (*side-note: i do like that there is a way to directly write **only** the array node, but i think that result is currently a little conflated and unclear*)
2. I think `downsampling_method` should only determine the downsampling method, and possibly *whether* or not downsampling occurs (its presence should *also* not be used to gate whether you get an array node or multiscales group)
3. `chunk_size_px` should not be used to determine how *many* downscaling levels you get (sure, it can determine what the maximum valid number of scales could be, but you should be able to select number of downscales without modifying the chunk layout of the top level array).

-----

>[!Note]
> _Background_:  generally speaking, I think **most typical experiments fall into one of three categories**  
> (at least as far as what ome-zarr can currently represent)
> 1. A **single ≤5D image** (TCZYX, at a single position).  Whether it has multiple scales or not is optional.
> 2. An **HCS plate**, where each FOV is a single ≤5D image
> 3. A **collection of ≤5D images**, usually corresponding to different positions.   There's no great spec for this at the moment (there's an old ongoing discussion https://github.com/ome/ngff/issues/31), but the best option at the moment is piggy backing on the bioformats2raw.layout transitional spec

This issue is mostly about 3 (collections), but touches on 1 (multiscales) images.

Generating a valid OME multiscales image for a single position is easy enough.  Just use the standard pattern, but *make sure not to provide output_key*.  This works (though it's semantically a bit confusing/magic ... and that's what #181 is discussing)

```python
settings = StreamSettings(
    arrays=[
        ArraySettings(
            output_key="",  # any other value will NOT result in an OME multiscales group
            ...
        )
    ]
)
```

However, **generating a collection of multiscale images is currently not possible, unless you *also* opt-in to downsampling**. 

This will NOT work... You will get `root.zarr/Pos0` and `root.zarr/Pos1`, but providing `output_key` means that each of these is directly an Array node, no multiscales metadata is written.

```python
settings = StreamSettings(
    arrays=[
        ArraySettings(
            output_key="Pos0",
            ...
        ),
        ArraySettings(
            output_key="Pos1",
            ...
        )
    ]
)
```

This will work, but it's a hack (the presence of `downsampling_method` shouldn't be related to group structure), and there's no way to say "i only want a single scale"... the number of downsampled scales is always automagically calculated from the chunking scheme.

```python
settings = StreamSettings(
      arrays=[
          ArraySettings(
              output_key="Pos0",
              downsampling_method=DownsamplingMethod.MEAN,
              ...
          ),
          ArraySettings(
              output_key="Pos1",
              downsampling_method=DownsamplingMethod.MEAN,
              ...
          )
      ]
  )
```

### LINKED ISSUE #252: enable downsampling to be optional
max_levels has no "zero additional levels" value; 0 is overloaded as no-limit. Proposal: 0 means zero additional and -1/None means no limit, or add an explicit multiscale: bool.

aliddell:
Closes #182 

Previously, OME-NGFF multiscales wrapping was implicitly controlled by the
`multiscale` boolean on `ZarrArraySettings`, which was itself tied to whether
a downsampling method was set. This conflated two distinct concerns: whether
to generate a downsampled image pyramid, and whether to wrap the array in an
OME-NGFF multiscales group.

This PR decouples those concerns by:

- Replacing `ZarrArraySettings.multiscale` with `is_ngff`
- Adding `ZarrDownsamplingMethod_None` as a zero-value sentinel, making
  `ZarrDownsamplingMethod_Decimate` no longer the implicit default
- Moving `make_array()` from `array.base.cpp` into `zarr.stream.cpp`, where
  it belongs as a factory driven by stream-level configuration
- Propagating `is_ngff` through `ArrayConfig` so array construction no longer
  depends on a separate `is_hcs_array` flag through the call chain

**Resulting behavior:**

| `downsampling_method` | Array belongs to `FieldOfView` | `is_ngff` | Result |
|---|---|---|---|
| `None` | No | `False` | Simple array node |
| `None` | No | `True` | Single-level OME-NGFF multiscales group, no image pyramid |
| `None` | Yes | (coerced) | Single-level OME-NGFF multiscales group, no image pyramid |
| set | No or Yes | (coerced) | OME-NGFF multiscales group with image pyramid starting at level `0` |

Python bindings expose `is_ngff` as a settable property with the same
coercion semantics, covered by `test_is_ngff_coercion`. The `test_ngff_streams`
test verifies end-to-end behavior for both the `is_ngff`-only and
`is_ngff` + downsampling paths.


CHANGELOG.md
```diff
@@ -9,9 +9,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+- `is_ngff` flag on `ZarrArraySettings` to explicitly request OME-NGFF multiscales wrapping without
+  downsampling (#213)
+- `ZarrDownsamplingMethod_None` sentinel value for explicitly representing "no downsampling" (#213)
 - `ZARR_DIRECT_IO` environment variable: when set, file handles are opened with `O_DIRECT` so writes bypass the OS page
   cache. Off by default, Linux only, and only valid on filesystems that accept unaligned direct writes, such as NFS
 
+### Changed
+
+- `ZarrArraySettings.multiscale` has been replaced by `is_ngff`; setting a downsampling method continues to
```
jeskesen: It might it be more accurate to have a "### Removed" section, and say that the paramter has been removed and the functionality has been replaced by setting the downsampling method to ZarrDownsamplingMethod_None?

README.md
```diff
@@ -453,7 +454,8 @@ flushed to disk when the stream is closed.
 
 ### High-content screening workflows
 
-The library supports high-content screening (HCS) datasets following the [OME-NGFF 0.5](https://ngff.openmicroscopy.org/0.5/) specification.
+The library supports high-content screening (HCS) datasets following
+the [OME-NGFF 0.5](https://ngff.openmicroscopy.org/0.5/) specification.
```
jeskesen:
```suggestion
the [OME-NGFF 0.5](https://ngff.openmicroscopy.org/0.5/) (Next-Generation File Format) specification.
```

CHANGELOG.md
```diff
@@ -9,9 +9,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `is_ngff` flag on `ZarrArraySettings` to explicitly request OME-NGFF multiscales wrapping without
+  downsampling (#213)
```
jeskesen:
```suggestion
  requiring downsampling (#213)
```

include/zarr.types.h
```diff
@@ -89,11 +89,19 @@ extern "C"
 
     typedef enum
     {
-        ZarrDownsamplingMethod_Decimate = 0,
-        ZarrDownsamplingMethod_Mean,
-        ZarrDownsamplingMethod_Min,
-        ZarrDownsamplingMethod_Max,
-        ZarrDownsamplingMethodCount,
+        ZarrDownsamplingMethod_None = 0, /**< No downsampling. When used with
```
jeskesen, line 98: This breaks enum compatibility...  not horrible, but I don't see it noted anywhere.  Is this purely internal, or could any users be setting this explicitly in configs?

python/acquire-zarr-py.cpp
```diff
@@ -1914,12 +1935,16 @@ PYBIND11_MODULE(acquire_zarr, m)
                        case ZarrDownsamplingMethod_Max:
                            method_str = "DownsamplingMethod.MAX";
                            break;
+                       case ZarrDownsamplingMethod_None:
```
jeskesen, line 1938: Should there be a DownsamplingMethod.NONE?

python/acquire_zarr/__init__.pyi
```diff
@@ -295,13 +308,13 @@ class DownsamplingMethod:
 
     DECIMATE: ClassVar[
```
jeskesen, line 314: Could the shifted enum values be a problem pertaining to previously pickled parameters.  (the allitererations sometimes write themselves)

As mentioned elsewhere, I think it might be good to register NONE as a value here.

src/streaming/settings.io.cpp
```diff
@@ -202,6 +202,7 @@ constexpr EnumEntry kCodecs[] = {
 };
 
 constexpr EnumEntry kDownsamplingMethods[] = {
+    { "none", ZarrDownsamplingMethod_None },
```
jeskesen, line 209: Do you ever bump schema versions when these values change?

jeskesen (approved):
I've put concerns regarding breaking changes inline above, and also some documentation nits/suggestions.   Nothing that should hold up the approval though.  Feel free to proceed as you see fit.

CHANGELOG.md
aliddell: Agreed, more accurate.

README.md
aliddell: Applied, thanks

CHANGELOG.md
aliddell: Applied, thanks

include/zarr.types.h
aliddell: It's internal. Configs don't use integers for this, but good to have a message in the changelog.

python/acquire-zarr-py.cpp
aliddell: This is deliberate. We already use Python's `None` here, having a duplicate option would be a belt-n-suspenders approach I'd like to avoid.

python/acquire_zarr/__init__.pyi
aliddell: Yeah, good point. No way to fix this while also making `None` the zero value, so we have to note it as breaking in the changelog.

src/streaming/settings.io.cpp
aliddell: First change, good catch.

python/acquire_zarr/__init__.pyi
jeskesen: Does none HAVE to be zero?  It would simplify everything mechanically in this PR if it weren't.  However, it wouldn't be intuitive.

PR #219, opened 2026-04-14
Restore compress/write parallelism

### LINKED ISSUE #218: Restore parallel compression and shard write
Opening an issue here for our previous discussion, see https://github.com/pymmcore-plus/pymmcore-plus/issues/611

aliddell:
Fixes #218.

## Write shard layers in parallel

This branch refactors the streaming pipeline to write chunk layers concurrently across worker threads. Chunk data is moved into the thread worker for writing as soon as all frames for that chunk are written, allowing appending to continue even as frames are written.

### Key changes

**New `Chunk` and `Shard` types** (`chunk.hh/cpp`, `shard.hh/cpp`)

`Chunk` is a self-contained unit owning its buffer, per-tile writes, and optional in-place compression. It skips writing all-zero tiles, avoiding unnecessary data in sparse chunks. `Shard` owns its sink and tracks chunk offset/extent tables, writing the checksummed shard index atomically when all its chunks have been accounted for. These replace the previous inline shard-table management scattered through `Array`.

**Frame ordering enforcement**

`FrameQueue::push/pop` now carry a `frame_id`. `Array::write_frame` checks that the incoming frame ID matches `frames_written()` and returns the new `WriteResult::FrameOutOfOrder` if a frame arrives early (so the caller can retry). ZarrStream` stamps each enqueued frame with a per-array `frames_queued` counter.

**Thread pool retry capability**

`ThreadPool::Task` return type changed from `bool` to `TaskResult` (`Success` / `Retry` / `Fatal`). Worker threads requeue `Retry` jobs up to a configurable limit rather than failing immediately. `Fatal` drains and stops the pool. Two new static helpers — `execute_job` and `execute_job_with_retry` — allow synchronous use without pushing to the queue.

**Remove `LockedBuffer`**

`LockedBuffer` is gone. `FrameQueue` uses `std::vector<uint8_t>` internally; `Array` uses `shared_ptr<Chunk>` with per-chunk `std::mutex` entries for tile writes under OpenMP.

**`FileHandlePool` becomes an LRU cache**

The file handle pool evicts the least-recently-used handle when the OS limit is reached, replacing the previous fixed capacity approach.

**Benchmark tooling rewrite**

`benchmarks/benchmark.py` and the new `benchmarks/main.py` significantly expand the benchmark harness.

### Testing

All existing unit and integration tests updated to the new `TaskResult`/`frame_id` signatures.


src/streaming/shard.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+    bool res = sink_->write(offset, buffer);
+    if (!res) {
+        LOG_ERROR("Failed to write chunk");
+        return false;
+    }
+
+    std::unique_lock lock(mutex_);
+    CHECK(unwritten_chunks_ > 0);
+    if (unwritten_chunks_.fetch_sub(1) == 0) {
```
nclack: Looks like this might always be false. I believe `fetch_sub(1)` returns the prior value, not the value after subtracting 1.

src/streaming/thread.pool.cpp
```diff
@@ -100,9 +146,23 @@ zarr::ThreadPool::process_tasks_()
...
+
+            switch (std::string err_msg; job->task(err_msg)) {
+                case TaskResult::Success:
+                    break;
+                case TaskResult::Retry:
+                    if (const auto max_retries =
+                          job->max_retries ? *job->max_retries : 3;
+                        job->attempt < max_retries) {
+                        ++job->attempt;
+                        push_to_queue_(std::move(*job)); // requeue
```
nclack, line 159: this mutates the queue but outside of the mutex, which was unlocked just before the `switch`

src/streaming/thread.pool.cpp
```diff
@@ -100,9 +146,23 @@ zarr::ThreadPool::process_tasks_()
...
+                    }
+                    break;
+                case TaskResult::Fatal:
+                    error_messages_.push_back(err_msg);
+                    error_handler_(err_msg);
+                    accepting_jobs = false; // drain and stop
```
nclack, line 169: need a notify_all to trigger parked workers.

src/streaming/zarr.stream.cpp
nclack:
```suggestion
LOG_ERROR("Timed out waiting for frame queue to finalize after 300s; "
"proceeding with shutdown");
```

nclack (commented):
found a few things. Looks like there's a test against an atomic that needs to get fixed, and maybe a thread safety thing in the thread pool.

ieivanov:
Thanks @aliddell! Yes, I can test this if you can help me with the installation. I'm targeting Windows and python 3.11 - it would be great if you can build me a wheel

src/streaming/shard.cpp
```diff
@@ -0,0 +1,147 @@
...
+        LOG_ERROR("Failed to write chunk");
+        return false;
+    }
+
+    std::unique_lock lock(mutex_);
+    CHECK(unwritten_chunks_ > 0);
+
+    // fetch_sub returns the value immediately preceding mutation
+    if (unwritten_chunks_.fetch_sub(1) == 1) {
+        res = write_table_();
```
nclack, line 88: need to set table_flushed_?

src/streaming/thread.pool.cpp
```diff
@@ -100,9 +146,25 @@ zarr::ThreadPool::process_tasks_()
...
+                    break;
+                case TaskResult::Retry:
+                    if (const auto max_retries =
+                          job->max_retries ? *job->max_retries : 3;
+                        job->attempt < max_retries) {
+                        ++job->attempt;
+                        lock.lock();
+                        push_to_queue_(std::move(*job)); // requeue
+                    }
+                    break;
```
nclack: should log and probably fail if max_retries is hit, since at this point you just lose data.

src/streaming/shard.cpp
```diff
@@ -0,0 +1,147 @@
...
+        return false;
+    }
+
+    uint64_t offset;
+    extents_[internal_index] = buffer.size();
+
+    {
+        std::unique_lock lock(mutex_);
+        offset = offsets_[internal_index] = file_offset_;
+        file_offset_ += buffer.size();
```
nclack, line 74: bumping file_offset_ here, but if write fails later, then this has the wrong value. My guess is this might be by design? But might need a comment.

src/streaming/chunk.cpp
```diff
@@ -0,0 +1,99 @@
...
+{
+    return size_bytes_;
+}
+
+bool
+zarr::Chunk::compress_and_take_buffer(
+  const std::optional<CompressionParams>& compression_params,
+  std::vector<uint8_t>& data)
+{
+    if (is_compressed_ || !compression_params) {
```
nclack: checking the condition outside the lock looks racy

src/streaming/thread.pool.cpp
```diff
@@ -69,7 +108,14 @@ zarr::ThreadPool::n_threads() const
     return threads_.size();
 }
 
-std::optional<zarr::ThreadPool::Task>
+void
+zarr::ThreadPool::push_to_queue_(TaskWrapper&& job)
+{
+    jobs_.push(job);
```
nclack:
```suggestion
    jobs_.push(std::move(job));
```

src/streaming/file.handle.cpp
```diff
@@ -33,32 +41,77 @@ zarr::FileHandle::get() const
...
 {
     std::unique_lock lock(mutex_);
-    if (n_active_handles_ >= max_active_handles_) {
-        cv_.wait(lock,
-                 [this]() { return n_active_handles_ < max_active_handles_; });
+
+    // block until we can either serve from cache or evict something
+    cv_.wait(lock, [&] {
+        return cache_.contains(filename) ||
+               cache_.size() < max_active_handles_ || evict_lru_();
```
nclack: evict_lru in the wait predicate is heavy. Can this be moved off the predicate? The problem is that evict ends up doing something with a FileHandle - maybe there's a predicate on the lru (is_full) or something that should be used instead.

src/streaming/shard.cpp
aliddell: I do this now in `write_table_()` and fixed a bug in the destructor while I was at it.

src/streaming/shard.cpp
aliddell: Commented

ieivanov:
Hi Alan! I had a chance to test this yesterday. I found that it works well - I no longer run into OOM error as described in https://github.com/pymmcore-plus/pymmcore-plus/issues/611 and my CPU stays at ~100% utilization for the duration of data writing (before there were clear dips). 

However, I also found that this version is ~10% slower compared to v6.0.0. Do you see the same in your benchmarks? Could this be an easy fix? I'm OK if you merge this PR without resolving the speed issue - it's up to you. Thanks again!

nclack (approved):

ieivanov:
Yay! I'll give this latest version a try and I'll let you know if anything looks amiss. Thanks for your help with this.

PR #222, opened 2026-04-18
Fix worker-exit deadlock in ZarrStream (#221)

### LINKED ISSUE #220: Flush errors swallowed — tests pass while data integrity fails
## Summary

Internal flush failures from `compress_and_flush_data_` are logged but not propagated back to the caller of `ZarrStream_append` on the call that triggered them. They only surface on a *subsequent* `append` via the `error_` member — and if no such call occurs, they never surface at all. The test passes even though an error was logged.

## Evidence

Seen in a recent CI run on `macos-latest` during `test_append_throws_on_overflow`:

```
[ERROR] array.cpp:913 compress_and_flush_data_: Failed to flush data:
  Failed to open file: '/private/var/.../test_append_throws_on_overflow.zarr/c/0/0/0':
  No such file or directory
python/tests/test_stream.py::test_append_throws_on_overflow PASSED
```

The first `stream.append(data)` had an internal flush failure; its zarr chunk `c/0/0/0` was never written. The test still passed because `pytest.raises(RuntimeError)` only verifies the *second* (deliberately-overflowing) append throws.

## Root cause (sketch)

`src/streaming/array.cpp:910-914` runs the flush job either on the thread pool or inline. On failure it does `LOG_ERROR(err); all_successful.fetch_and(success)` and `promise->set_value()`, but the `all_successful` result only lands in `error_` asynchronously via the frame-queue thread. The append that *triggered* the flush has already returned `Success` by then.

## Proposed fix

Surface flush failures synchronously on the triggering call — wait on outstanding futures and convert to an error return before `ZarrStream_append` returns. Alternatively, at minimum, surface them at `ZarrStream_destroy`/`close` time so a `close` after a failed flush raises.

## Related

- `python/tests/test_stream.py::test_append_throws_on_overflow` — the test that surfaced this.

### LINKED ISSUE #221: ZarrStream worker error path can deadlock append() and the destructor
## Summary

If the frame-processor worker thread takes either of its early-exit error paths in `src/streaming/zarr.stream.cpp` (lines `1657-1660` and `1669-1673`), it leaves shared state inconsistent in a way that can deadlock both `ZarrStream::append()` and the destructor's `finalize_frame_queue_()`. The worker only notifies one of the three condition variables that other threads wait on, and never sets `process_frames_ = false`.

## Where

`src/streaming/zarr.stream.cpp`, `process_frame_queue_()` worker — both early-exit paths look like:

```cpp
set_error_("Output node not found for key: '" + output_key + "'");
std::unique_lock lock(frame_queue_mutex_);
frame_queue_finished_cv_.notify_all();
return;
```
(also the `write_frame` failure path immediately below)

`set_error_` (line `1429`) only sets `error_`. It does not touch `process_frames_` and notifies nothing.

### Scenario A — `append()` blocks indefinitely

`zarr.stream.cpp:957-960` and `:981-983`:

```cpp
std::unique_lock lock(frame_queue_mutex_);
while (!frame_queue_->push(frame_buffer, key) && process_frames_) {
    frame_queue_not_full_cv_.wait(lock);   // unbounded, no timeout
}
```

If the worker exited via the early-return path, `process_frames_` is still `true` and the queue stays full. `append()` is blocked on `frame_queue_not_full_cv_`. The worker only notified `frame_queue_finished_cv_` on its way out — different CV, no wakeup. Wait is unbounded, never returns.

### Scenario B — destructor blocks indefinitely

`zarr.stream.cpp:1714-1715`:

```cpp
std::unique_lock lock(frame_queue_mutex_);
frame_queue_finished_cv_.wait(lock,
    [this] { return frame_queue_->empty(); });
```

If the worker took an early return with frames still queued, the predicate `frame_queue_->empty()` stays `false`. The cleanup at `1689-1693` that clears the queue lives on the **normal-exit** path of the worker, which the early returns skip. Destructor wakes once on the worker's `notify_all`, sees the predicate is still false, and goes back to sleep forever.

## How it surfaces in CI

The macOS log captured in our recent CI run for issue #220 (`array.cpp:913` `compress_and_flush_data_: Failed to flush data: ... No such file or directory`) is exactly the kind of failure that flows up into `write_frame` returning non-OK and the worker hitting line `1672`. So this bug and #220 are connected: a swallowed flush error elsewhere produces the worker-exit path that triggers the deadlock here.

We see this most often on Windows, presumably because Windows' file-handle and lock semantics make the underlying chunk-write failures more frequent across rapid pytest setup/teardown, but the deadlock itself is a pure data-race, not Windows-specific. Concretely: `python/tests/test_stream.py::test_anisotropic_downsampling` in our `Test Python on windows-latest` job hangs at `stream.append(data)` past the pytest-timeout — pytest can't preempt the C call because the wait is unbounded inside the C library.

## Proposed fix

In both early-exit paths, set `process_frames_ = false`, drain the queue, and notify all the relevant CVs while holding `frame_queue_mutex_`:

```cpp
set_error_("...");
{
    std::unique_lock lock(frame_queue_mutex_);
    process_frames_ = false;
    frame_queue_->clear();
    frame_queue_not_full_cv_.notify_all();   // unblock append()
    frame_queue_empty_cv_.notify_all();
    frame_queue_finished_cv_.notify_all();   // unblock destructor wait
}
return;
```

Belt-and-suspenders: at `:1714`, replace `wait` with `wait_for(timeout, predicate)`, log if the timeout expires. The destructor should never be able to hang the process indefinitely even if a future code change reintroduces a missing notify.

## Related

- #220 — flush errors logged but swallowed in `compress_and_flush_data_`. The flow there is one of the producers of `write_frame` failures that trigger the deadlock above.

nclack:
## Summary

The two early-exit paths in `ZarrStream_s::process_frame_queue_()` (missing output array, failed `write_frame`) left `process_frames_ = true`, did not drain the queue, and only notified `frame_queue_finished_cv_`. A producer blocked in `append()` on `frame_queue_not_full_cv_` (and the destructor, blocked on the same condition) therefore deadlocked once the worker returned. Fixes #221.

On each early-exit path the worker now, under `frame_queue_mutex_`:
- sets `process_frames_ = false`
- clears the queue
- notifies all three relevant CVs (`frame_queue_not_full_cv_`, `frame_queue_empty_cv_`, `frame_queue_finished_cv_`)

`finalize_frame_queue_()` now uses `wait_for(300s, predicate)` and logs an error on timeout, so a future regression in the worker cannot hang shutdown indefinitely.

Also wraps the `write_frame()` call in a `try`/`catch` so that a `CHECK()` throw from a failing `compress_and_flush_data_` (the issue described in #220) is captured and surfaced via `set_error_()` instead of propagating out of the worker thread and calling `std::terminate`. The error then becomes visible to the next `append()` via the existing `if (!error_.empty())` guard at the top of `append()`. Fixes #220.

## Test plan

- [ ] add a regression test that forces a write failure and verifies neither `append()` nor the destructor hangs

### DIFF
--- a/src/streaming/zarr.stream.cpp
+++ b/src/streaming/zarr.stream.cpp
@@ -1656,21 +1656,49 @@ ZarrStream_s::process_frame_queue_()
             it == output_arrays_.end()) {
             // If we have gotten here, something has gone seriously wrong
             set_error_("Output node not found for key: '" + output_key + "'");
-            std::unique_lock lock(frame_queue_mutex_);
-            frame_queue_finished_cv_.notify_all();
+            {
+                std::unique_lock lock(frame_queue_mutex_);
+                process_frames_ = false;
+                frame_queue_->clear();
+                frame_queue_not_full_cv_.notify_all();
+                frame_queue_empty_cv_.notify_all();
+                frame_queue_finished_cv_.notify_all();
+            }
             return;
         } else {
             auto& output_node = it->second;
 
             size_t n_bytes;
-            if (const auto result =
+            bool write_ok = false;
+            std::string err_msg;
+            try {
+                const auto result =
                   output_node.array->write_frame(frame, n_bytes);
-                result != zarr::WriteResult::Ok) {
-                // TODO (aliddell): retry on WriteResult::PartialWrite
-                set_error_("Failed to write frame to writer for key: " +
-                           output_key);
-                std::unique_lock lock(frame_queue_mutex_);
-                frame_queue_finished_cv_.notify_all();
+                if (result != zarr::WriteResult::Ok) {
+                    // TODO (aliddell): retry on WriteResult::PartialWrite
+                    err_msg = "Failed to write frame to writer for key: " +
+                              output_key;
+                } else {
+                    write_ok = true;
+                }
+            } catch (const std::exception& exc) {
+                // A failed compress/flush inside write_frame throws via
+                // CHECK(); surface it back to the user so the next append()
+                // sees the error instead of the failure being swallowed.
+                err_msg = "Exception while writing frame for key '" +
+                          output_key + "': " + exc.what();
+            }
+
+            if (!write_ok) {
+                set_error_(err_msg);
+                {
+                    std::unique_lock lock(frame_queue_mutex_);
+                    process_frames_ = false;
+                    frame_queue_->clear();
+                    frame_queue_not_full_cv_.notify_all();
+                    frame_queue_empty_cv_.notify_all();
+                    frame_queue_finished_cv_.notify_all();
+                }
                 return;
             }
         }
@@ -1710,11 +1738,18 @@ ZarrStream_s::finalize_frame_queue_()
         frame_queue_not_full_cv_.notify_all();
     }
 
-    // Wait for frame processing to complete
+    // Wait for frame processing to complete. Use a timeout so the destructor
+    // cannot hang indefinitely if a future change reintroduces a missing
+    // notify on one of the worker-thread early-exit paths.
     std::unique_lock lock(frame_queue_mutex_);
-    frame_queue_finished_cv_.wait(lock, [this] {
-        return frame_queue_processing_done_.load() || frame_queue_->empty();
-    });
+    if (!frame_queue_finished_cv_.wait_for(
+          lock, std::chrono::seconds(30), [this] {
+              return frame_queue_processing_done_.load() ||
+                     frame_queue_->empty();
+          })) {
+        LOG_ERROR("Timed out waiting for frame queue to finalize after 30s; "
+                  "proceeding with shutdown");
+    }
 }
 
 bool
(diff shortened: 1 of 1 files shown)


src/streaming/zarr.stream.cpp
```diff
@@ -1710,11 +1738,18 @@ ZarrStream_s::finalize_frame_queue_()
...
-    // Wait for frame processing to complete
+    // Wait for frame processing to complete. Use a timeout so the destructor
+    // cannot hang indefinitely if a future change reintroduces a missing
+    // notify on one of the worker-thread early-exit paths.
     std::unique_lock lock(frame_queue_mutex_);
-    frame_queue_finished_cv_.wait(lock, [this] {
-        return frame_queue_processing_done_.load() || frame_queue_->empty();
-    });
+    if (!frame_queue_finished_cv_.wait_for(
+          lock, std::chrono::seconds(30), [this] {
```
nclack:
```suggestion
          lock, std::chrono::seconds(300), [this] {
```

nclack:
cherry-picked this over to #212

aliddell (approved):

PR #231, opened 2026-06-10
Propagate shard flush failures instead of swallowing them (#229)

### LINKED ISSUE #229: flush_file I/O errors on NFS are silently swallowed, producing corrupt output with exit code 0
Problem

  When writing to an NFS/Lustre filesystem under high concurrency, flush_file in platform.cpp can fail with Input/output
   error. The error is logged but not raised so the write returns success to the caller, the process exits 0, and the
  resulting shard files are silently corrupt.

  Log output observed:
  platform.cpp:93 flush_file: Failed to flush file: Input/output error

  Reading back the written shards then raises:
  RuntimeError: error during blosc decompression: 0

  The corruption is partial and timepoint-dependent: shards written during peak NFS contention are corrupt while earlier
   ones are fine. This makes it extremely hard to detect without reading back every chunk.

  Environment: NFS/Lustre archive filesystem, 200+ parallel writers, acquire-zarr 0.8.0

  Expected behavior
  If flush_file fails, the error should propagate up as a C++ exception (and surface as a Python exception), so the
  caller knows the write failed and can exit non-zero, retry, or abort cleanly.

  Suggested fixes
  1. Propagate the error: Raise from flush_file rather than logging and continuing, so callers always know when a write
  has failed.
  2. Atomic shard writes: Write each shard to a temp file and rename it into place only on success. This prevents a
  partially-flushed file from ever appearing as a valid (but corrupt) shard, regardless of whether the exception
  propagates correctly.

  Reproduction

  The failure is not easily reproducible without an NFS filesystem under write contention, but the pattern is:

```
  import acquire_zarr as aqz
  import numpy as np
  import zarr

  settings = aqz.StreamSettings(
      store_path="/nfs/archive/test.zarr",  # NFS path under high concurrency
      arrays=[
          aqz.ArraySettings(
              output_key="0",
              shape=(1, 1, 1, 2048, 2048),
              chunks=(1, 1, 1, 2048, 2048),
              dtype=aqz.DataType.UINT16,
              version=aqz.ZarrVersion.V3,
          )
      ],
      version=aqz.ZarrVersion.V3,
  )

  stream = aqz.ZarrStream(settings)
  for t in range(141):
      block = np.random.randint(0, 65535, (1, 1, 1, 2048, 2048), dtype=np.uint16)
      stream.append(block, key="")
```

  stream.close()  # Returns without raising, even when flush failed.
                  # stderr shows: platform.cpp:93 flush_file: Failed to flush file: Input/output error

  Reading back the written data then raises:
  arr = zarr.open("/nfs/archive/test.zarr/0", mode="r")
  np.asarray(arr[70, 0, 0, :4, :4]) # RuntimeError: error during blosc decompression: 0

aliddell:
Fixes #229.

## Problem

When writing to NFS/Lustre under high concurrency, a failed `flush_file` (fsync) was logged but never raised: the write reported success, the process exited 0, and the resulting shards were silently corrupt (`RuntimeError: error during blosc decompression` on read-back).

The root cause is structural: the only `fsync` in the write path ran exclusively inside `~Shard()`, via `finalize_sink(std::move(sink_))` whose return value was discarded — and a destructor can't throw anyway. On NFS the `pwrite`s succeed into the page cache; the real write-back error only surfaces at `fsync`. Two further links also dropped the signal: `finalize_stream` didn't check `error_`, and `ZarrStream_destroy` returns `void`.

## Changes

**Move shard finalization out of the destructor and onto checked paths**
- `Shard::finalize()` / `finalize_unlocked_()` (idempotent): write the table if needed, flush the sink (fsync / S3 multipart completion), return the result.
- The **last chunk writer** now finalizes the shard, so a flush failure becomes that worker job's `TaskResult::Fatal` → `set_error_`, surfaced on the next `append` and at close. This covers the rollover/streaming case behind the timepoint-dependent corruption.
- `~Shard()` keeps `finalize()` as a best-effort, log-only fallback for aborted streams.

**Propagate at close**
- `Array::finalize_shards_()` explicitly finalizes the partial trailing shard once writes drain; `Array::close_()` fails if it does.
- `finalize_stream` returns `false` when `error_` is set (catches async job failures).

**Surface across the C/Python boundary**
- New additive `ZarrStatusCode ZarrStream_close(ZarrStream*)`; `ZarrStream_destroy` is now a void wrapper around it (non-breaking).
- Python `close()` calls `ZarrStream_close` and raises `RuntimeError` on a non-success status.

## Tests

`tests/unit-tests/shard-finalize.cpp` covers last-writer finalization, idempotency, and explicit finalize of a partial shard.

Note: the test guards the refactor but does not force an `fsync` failure — that isn't portably/deterministically reproducible (read-only-dir injection is bypassed under root). The failure path is wired through return values existing tests exercise; a sink-injection seam could be added for hard coverage of the error path if desired.

## Not included (deferred)

Atomic temp-file + rename for shards (issue suggestion #2) is left for a separate PR.


src/streaming/shard.cpp
```diff
@@ -81,11 +84,12 @@ zarr::Shard::write_chunk(uint32_t internal_index,
...
     CHECK(unwritten_chunks_ > 0);
 
-    // last writer in the shard publishes the table
+    // last writer in the shard publishes the table and durably flushes the
+    // shard, so a flush failure surfaces as this job's result
     bool res = true;
     if (unwritten_chunks_.fetch_sub(1) == 1) {
         std::unique_lock lock(mutex_);
-        res = write_table_();
+        res = finalize_unlocked_();
```
nclack, line 103: This return value now conflates a retryable chunk-write failure with a non-retryable finalize failure, and `try_write` retries both:

- table write fails but flush succeeds → `finalize_sink` resets `sink_` → the retry null-derefs at `sink_->write`
- flush fails (the #229 case) → the retry only survives by tripping `CHECK(unwritten_chunks_ > 0)`, masking the real error

Suggest guarding re-entry in `write_chunk`/`skip_chunk`:
```cpp
{
    std::unique_lock lock(mutex_);
    if (finalized_) {
        return finalize_ok_;
    }
}
```
Retries then exhaust into the existing `retries_exhausted_msg()` → Fatal path. A unit test calling `write_chunk` after `finalize()` would pin this.

src/streaming/zarr.stream.cpp
```diff
@@ -1781,6 +1781,13 @@ finalize_stream(ZarrStream* stream)
     // thread
     stream->thread_pool_->await_stop();
 
+    // a worker job (e.g. a shard flush during streaming) may have failed
+    // asynchronously; surface it rather than reporting a clean close
+    if (!stream->error_.empty()) {
```
nclack, line 1788: This misses Fatal jobs that never set `err`: the empty-chunks `skip_chunk` job in `compress_and_flush_data_` leaves `err` unset on failure → `error_handler_("")` → `error_` stays empty. If that shard is then dropped at rollover, its cached failure dies with it — silent again. Set `err` there, and consider having `set_error_` substitute a generic message for an empty one.

src/streaming/acquire.zarr.cpp
```diff
@@ -615,13 +615,23 @@ extern "C"
...
-        if (!finalize_stream(stream)) {
-            return;
+        if (stream == nullptr) {
+            return ZarrStatusCode_Success;
         }
 
+        const bool ok = finalize_stream(stream);
         delete stream;
+
+        return ok ? ZarrStatusCode_Success : ZarrStatusCode_InternalError;
```
nclack: nit: `ZarrStatusCode_IOError` is more apt than `InternalError` for a failed flush.

python/acquire-zarr-py.cpp
```diff
@@ -1309,14 +1309,25 @@ class PyZarrStream
             return;
         }
 
+        ZarrStatusCode status = ZarrStatusCode_Success;
         try {
-            stream_.reset(); // calls ZarrStream_destroy
+            // release ownership so the deleter does not also destroy the
+            // stream; ZarrStream_close finalizes, frees, and reports a failed
+            // flush (e.g. an I/O error on a network filesystem)
+            status = ZarrStream_close(stream_.release());
```
nclack: nit (pre-existing, but easy now): this blocks on the queue drain + fsyncs while holding the GIL; wrap the call in `py::gil_scoped_release` — the error handling already happens after it.

nclack (changes requested):
Architecture is right and this fixes #229 end-to-end. One crash path on the new error route should be fixed before merge (see `shard.cpp` comment); the rest are non-blocking:

- The inline-fallback paths (`execute_job*` when `push_job` fails or the pool has one thread) still only log, so a mid-stream rollover can still drop a failed shard silently. Worth recording in #229 alongside the atomic-rename follow-up.
- `Shard::make_sink_` doesn't null-check `make_file_sink`/`make_s3_sink`; under the new contract a sink-less shard "finalizes successfully" (`finalize_sink(nullptr)` returns true). A `CHECK(sink_)` in the ctor would fit.
- `shard.hh` lost its trailing newline.

jeskesen (approved):
LGTM.  This is timely for me, as we could try writing directly to NFS on my project.

PR #230, opened 2026-06-08
Streaming write-path performance improvements

aliddell:
## Streaming write-path performance improvements

Four focused commits on the local-filesystem write path. ~1.3–1.9× throughput across regimes and ~3× lower peak RSS. All non-S3 tests pass (34/34).

### Changes
- **build:** enable AVX2 + LTO for the streaming library
- **memory:** bound the frame queue to 256 MiB
- **win32:** reuse the per-handle OVERLAPPED event, drop per-close `FlushFileBuffers`
- **array:** copy frames directly into chunk buffers

### Results
Benchmark: 1024×1024 `uint16`, 512 frames (1 GiB), random data (worst case for compressors). `perf/bench_stream.py` scorecard, pre-patch vs full branch.

| regime | throughput | peak RSS |
|--------|-----------|----------|
| std (s4, ct32) | 682 → 867 MB/s (+27%) | 1019 → 345 MiB (−66%) |
| manyfiles (s1) | 349 → 581 MB/s (+67%) | 1070 → 505 MiB (−53%) |
| rollover (ct1) | 135 → 252 MB/s (+86%) | 1022 → 376 MiB (−63%) |
| threads4       | 552 → 887 MB/s (+61%) | 1015 → 617 MiB (−39%) |
| lz4 (blosc)    | 457 → 661 MB/s (+45%) | 1066 → 426 MiB (−60%) |
| multiscale     | 278 → 401 MB/s (+44%) | 1058 → 335 MiB (−68%) |

### Tradeoff
Bounding the queue trades peak memory for tail latency: `append()` now blocks under backpressure instead of buffering, so per-frame p99 rises (std 7 → 14 ms; worst case `rollover` 7 → 57 ms). Mean/median throughput climbs everywhere; the cost is confined to the tail when the producer outruns the writers.

Issue #232, opened 2026-06-22
Streaming buffers full extent of inner dimensions before flushing — unbounded RAM with a large inner (non-append) dimension


jeskesen:
## Summary

When a Zarr array has an inner (non-append, non-spatial) dimension with a large array extent, the streaming writer buffers the **entire extent** of that dimension in RAM before flushing anything to disk. Peak memory grows linearly with the inner dimension's size rather than staying bounded by chunk/shard size, and nothing is written until that full sweep completes.

In our application — high-throughput image streaming with multiple concurrent stores, ~10k frames per store — this makes the natural `[t, z, y, x]` layout unwritable: RAM exceeds the host's capacity and no files appear until the store closes.

Observed on acquire-zarr 0.7.0; confirmed still present on `main` (0.8.0).

## Layout that triggers it

A 4-axis store `[t, z, y, x]`, where `z` is an inner dimension with a large array extent, `t=1`, frame size 1764x2660 uint16 (9.4 MB):

```
dimensions: [ t (append), z (inner), y, x ]
chunk_z = 64
```

## What we observe

Single stream, paced at ~55 fps, peak RAM vs. inner-dim (`z`) size:

| z (inner) | peak RAM | flushed to disk during run? |
|---|---|---|
| 256 | 3.6 GB | no — 0 files until close |
| 512 | 6.0 GB | no — 0 files until close |
| 1024 | 10.8 GB | no — 0 files until close |

4 concurrent streams:

| z (inner) | peak RAM | aggregate throughput | backlog drain at close |
|---|---|---|---|
| 512 | 23.8 GB | 1086 MB/s | 16.7 s |
| 1024 | 43.0 GB | 1086 MB/s | 16.7 s |

RAM grows linearly with the inner dimension, extrapolating to ~410 GB across 4 stores for our full workload — well over the host's 377 GB. Chunking or sharding the `z` axis does **not** change this, because the flush threshold is computed from `array_size_px`, not `chunk_size_px`.

## Root cause

`src/streaming/array.cpp::should_flush_()` (current `main`):

```cpp
size_t frames_before_flush = dims->final_dim().chunk_size_px;   // append axis: chunk size
for (auto i = 1; i < dims->ndims() - 2; ++i) {
    frames_before_flush *= dims->at(i).array_size_px;           // inner axes: FULL extent
}
return frames_written_() % frames_before_flush == 0;
```

[...]

PR #233, opened 2026-06-22
Flush large intermediate dimensions band-by-band to bound memory

aliddell:
## Problem

When the append dimension has a chunk size of 1 and there is at least one intermediate dimension (e.g. `[t, z, y, x]` where `z` is a large grid/position axis), the writer buffered the **entire inner volume** before its first flush — a flush only fired once a whole append-chunk layer had arrived. With a ~62k intermediate axis this exhausts memory.

Reported downstream in czbiohub-sf/livescreen-acquisition#210 (per-camera scanning store reshaped to `[t, grid, y, x]`, `grid` up to ~62.5k).

## Change

Flush and free chunk buffers **one band at a time** along the dimension just inside the append axis as the inner sweep completes them, so peak raw memory tracks a single band instead of the full inner volume.

- Chunk buffers are now allocated lazily and reclaimed per band (previously the full inner volume was allocated up front in the `Array` ctor).
- Shards already stream compressed chunks to disk and self-finalize via the `unwritten_chunks_` countdown, so shards spanning multiple bands (or finishing mid-sweep) finalize correctly with no extra buffering.
- The per-chunk write/skip job dispatch was extracted into helpers shared by the whole-layer and banded paths.

### Scope / fallbacks

- Banding requires **append chunk size == 1** and **no transposition** (otherwise inner chunks span multiple sweeps, or acquisition order ≠ storage order, so a band could be flushed before it is complete). In those cases the previous whole-layer flush is kept, plus a one-line warning when the resident working set is large.
- Only bands along the dimension immediately inside the append axis; a huge intermediate dim that is *not* dim-1 is out of scope (the reported case has it as dim-1).

### Memory estimate

`estimate_max_memory_usage` now reports the reduced (single-band) bound when banding applies, and is also corrected to model the frame queue as the actual 256 MiB / [16, 512]-frame bound (#230) rather than a flat 1 GiB.

## Tests

- `array-dimensions-courtesy-flush` (unit): band-math helpers across even/ragged/multi-intermediate/transposed/no-intermediate configs.
- `test_intermediate_dimension_courtesy_flush` (Python, parametrized even/ragged z, with a trailing partial timepoint): full data round-trip, band-based memory estimate, and a guard that live memory stays well below the full inner volume.
- Updated `estimate-memory-usage` (C++) and `test_estimate_max_memory_usage` (Python) for the corrected queue bound.

Existing coverage guards the unchanged paths: `array-write-ragged-internal-dim` (append chunk > 1), `stream-multi-frame-append`, and `stream-with-ragged-final-shard` (banding with a single band).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Flush incrementally along a large intermediate dimension: when the append dimension has a chunk size of 1, chunk
+  buffers are now flushed and freed one band at a time along the dimension just inside the append axis, so peak memory
+  tracks a single band rather than the whole inner volume. This makes large intermediate axes (e.g. a `[t, z, y, x]`
+  store with a ~62k `z`) feasible without running out of memory. `estimate_max_memory_usage` reflects the reduced bound
+  (czbiohub-sf/livescreen-acquisition#210)
 - Bound the frame queue to 256 MiB; `append()` now applies backpressure instead of buffering unboundedly, cutting
   peak memory ~3x at the cost of higher tail latency under sustained pressure (#230)
 - Copy frames directly into chunk buffers, removing an intermediate copy (~1.3-1.9x write throughput on filesystem) (#230)
@@ -22,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `estimate_max_memory_usage` now models the frame queue as the actual 256 MiB / [16, 512]-frame bound (#230) instead
+  of a flat 1 GiB, so the estimate matches real peak usage
 - Shard flush (`fsync`) failures are no longer swallowed in the `Shard` destructor; an I/O error now fails the
   stream instead of silently producing corrupt shards. Python `close()` raises on a failed flush (#231)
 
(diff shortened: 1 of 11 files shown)


jeskesen:
Downstream verification of #232 — this branch resolves it. Built from `61ef225` and re-ran our `[t, grid, y, x]` benchmarks (1764×2660 uint16 = 9.4 MB frames, `t` chunk = 1 so banding engages, paced ~55 fps/camera).

**Synthetic, acquire-zarr peak RAM before (0.7.0) vs this branch:**

| config | 0.7.0 | this branch |
|---|---|---|
| single-stream, inner z = 256 / 512 / 1024 | 3.6 / 6.0 / 10.8 GB (grows with z) | 2.18 / 2.18 / 2.19 GB (flat) |
| 4-stream, z=1024, chunk_z=64 | 43.0 GB | 8.5 GB |
| 4-stream, z=1024, chunk_z=4 | never flushes | 2.08 GB |
| flush during run | never (0 files until close) | yes |
| 4-stream throughput / close-drain | 1086 MB/s / 16.7 s | 2065 MB/s / 0.3–0.8 s |

Peak now tracks one band = `chunk_z × frame_bytes`, independent of the inner extent — exactly the intended behavior.

**Real 4-camera workload** (single timepoint, 62582 frames/camera, ~22 min, chunk_z=4 / shard_z=16, blosc-zstd): zero frame loss (`written=62582` on all 4 streams), stores 180–188 GB each with data flushing throughout, writers kept up at 55 fps the whole run, RSS bounded/plateaued. The same run on 0.7.0 immediately beforehand reproduced the original failure — stall after ~115 frames, stores holding only group metadata.

Thanks for the quick turnaround.

jeskesen (approved):

jeskesen (approved):
Looks good to me.

PR #234, opened 2026-06-24
Truncate metadata writes; allow concurrent reads on Windows

aliddell:
## Motivation

Two related filesystem issues, surfaced by a request to drag-and-drop an in-progress acquisition into napari on Windows.

1. **Files were locked against readers on Windows.** Handles were opened with `dwShareMode = FILE_SHARE_WRITE` only, so any process opening the store for reading (napari) failed with a sharing violation for the duration of the acquisition.
2. **Shrinking metadata left stale trailing bytes.** `zarr.json` is written as a single whole-file write at offset 0, but neither the Win32 (`OPEN_ALWAYS`) nor POSIX (`O_WRONLY|O_CREAT`) backend truncates. A rewrite shorter than the previous version — e.g. replacing a large custom-metadata blob with a smaller one across two acquisitions to the same store — leaves the tail of the old document past the end of the new one. The library's own reader (nlohmann `operator>>`) stops at the first complete value and tolerated it, but strict parsers like zarr-python (`json.loads`) reject the result.

## Changes

- **Windows share mode:** open filesystem handles with `FILE_SHARE_READ | FILE_SHARE_WRITE` so readers can open the store during acquisition. No write-path / throughput impact — share mode only governs concurrent opens.
- **Truncate-to-fit:** new `truncate_file` platform primitive (`SetFileInformationByHandle(FileEndOfFileInfo)` on Win32, which is file-pointer-independent and so correct for the `FILE_FLAG_OVERLAPPED` handle; `ftruncate` on POSIX) and a `truncate_to_fit` flag on `FileSink` that clamps the file to `offset + data.size()` after a successful write. It is a post-write truncate (not an open-mode flag) because handles are pooled and reused across rewrites. Enabled for the array/root metadata sink and the intermediate group/plate/well metadata sink; data sinks keep the default (no truncation) since they write incrementally at offsets.

## Tests

Adds `tests/unit-tests/metadata-shrink-stale-bytes.cpp`, which reproduces the shrinking-metadata case via two `overwrite=false` acquisitions to the same store and asserts the resulting `zarr.json` is strictly parseable. It fails without the truncate fix and passes with it.

### DIFF
--- a/src/streaming/file.sink.cpp
+++ b/src/streaming/file.sink.cpp
@@ -9,10 +9,15 @@ seek_and_write(void* handle, size_t offset, ConstByteSpan data);
 bool
 flush_file(void* handle);
 
+bool
+truncate_file(void* handle, size_t size);
+
 zarr::FileSink::FileSink(std::string_view filename,
-                         std::shared_ptr<FileHandlePool> file_handle_pool)
+                         std::shared_ptr<FileHandlePool> file_handle_pool,
+                         bool truncate_to_fit)
   : file_handle_pool_(file_handle_pool)
   , filename_(filename)
+  , truncate_to_fit_(truncate_to_fit)
 {
     EXPECT(file_handle_pool_ != nullptr, "File handle pool not provided.");
 }
@@ -33,6 +38,10 @@ zarr::FileSink::write(size_t offset, ConstByteSpan data)
     bool retval = false;
     try {
         retval = seek_and_write(borrowed.handle_->get(), offset, data);
+        if (retval && truncate_to_fit_) {
+            retval =
+              truncate_file(borrowed.handle_->get(), offset + data.size());
+        }
     } catch (const std::exception& exc) {
         LOG_ERROR("Failed to write to file ", filename_, ": ", exc.what());
     }
--- a/src/streaming/file.sink.cpp
+++ b/src/streaming/file.sink.cpp
@@ -29,6 +29,17 @@ zarr::FileSink::write(size_t offset, ConstByteSpan data)
         return true;
     }
 
+    // truncate_to_fit_ clamps the file to offset + data.size(), which is only
+    // correct for a whole-file write at 0; a later write at a lower offset would
+    // truncate away earlier content. Enforce the invariant at its use site.
+    if (truncate_to_fit_ && offset != 0) {
+        LOG_ERROR("Truncating sink for ",
+                  filename_,
+                  " requires writes at offset 0; got offset ",
+                  offset);
+        return false;
+    }
+
     const auto borrowed = file_handle_pool_->get_handle(filename_);
     if (borrowed.handle_ == nullptr) {
         LOG_ERROR("Failed to get file handle for ", filename_);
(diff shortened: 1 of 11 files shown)


src/streaming/file.sink.cpp
```diff
@@ -33,6 +38,10 @@ zarr::FileSink::write(size_t offset, ConstByteSpan data)
     bool retval = false;
     try {
         retval = seek_and_write(borrowed.handle_->get(), offset, data);
+        if (retval && truncate_to_fit_) {
```
jeskesen, line 52: Possible race condition?  Could a read happen between seek_and_write / truncate_to_fit_?  Surely, its improbable.. but perhaps there's a straighforwared way to make this atomic?

src/streaming/file.sink.hh
```diff
@@ -24,5 +25,9 @@ class FileSink : public Sink
     std::shared_ptr<FileHandlePool> file_handle_pool_;
 
     std::string filename_;
+
+    // Whole-file replacement: drop bytes past what was written so a shorter
+    // rewrite leaves no stale tail. Used for metadata (single write at 0).
```
jeskesen, line 30: The comment says "single write at 0" but there's no assert(offset == 0) in FileSink::write() when truncate_to_fit_ is true. Currently all callers satisfy it, but a future caller that stores the sink and issues two writes would silently corrupt the file with no diagnostic. A one-line assert is cheap and makes the invariant self-documenting at the enforcement point rather than just in a comment.

jeskesen (approved):
2 small comments, I don't consider them blocking, but maybe take a look at them at least before merging.

src/streaming/file.sink.hh
aliddell: Good call — added in 0776adc. `FileSink::write` now rejects a non-zero offset when `truncate_to_fit_` is set (logs and returns false rather than throwing, to match the function's existing error style). The corruption mode is real since `truncate_file` clamps to `offset + data.size()`, so this enforces the whole-file-at-0 invariant at the use site instead of just in the comment.

src/streaming/file.sink.cpp
aliddell: You''re right that the window exists, but the truncate doesn''t actually open a new one — an in-place metadata rewrite was never atomic to begin with (a reader could always catch a partial `seek_and_write`). And for the shrink case specifically, this is strictly better than before: previously the stale tail was *permanent* (always broken for a strict parser like zarr-python), whereas now the final state is correct and only the tiny window between the write and the truncate is observable.

The only way to make it truly atomic is write-to-temp + atomic rename (`MoveFileEx` / `rename`). We considered that earlier and deferred it as more intrusive — and on Windows it has its own wrinkle: renaming over a `zarr.json` that a reader holds open without `FILE_SHARE_DELETE` (e.g. napari) can fail with a sharing violation, so it needs a retry loop. Given metadata writes are infrequent (rollover/close) and the reader here is an interactive drag-and-drop, I''d lean toward leaving it as-is and revisiting temp+rename only if torn metadata reads ever actually bite. Happy to file a follow-up issue to track it.

PR #236, opened 2026-06-25
Load and dump stream settings from YAML/JSON config files

aliddell:
## Summary

Adds config-file support so the declarative parts of `ZarrStreamSettings` can be expressed in a **YAML or JSON** file instead of being built entirely through the C/Python API. JSON is a subset of YAML, so a single **yaml-cpp** parse path handles both, bridged through `nlohmann::json`.

Covers: store path, `overwrite`, `max_threads`, S3 endpoint/bucket/region, arrays (data type, compression, multiscale, downsampling, `max_levels`, storage dimension order), dimensions (name/type/sizes/unit/scale), and full HCS plate/well/FOV/acquisition trees.

## API

**C** (`acquire.zarr.h`):
- `ZarrStreamSettings_load_from_file` / `_load_from_string` — owning; free with `ZarrStreamSettings_destroy_loaded`
- `ZarrStreamSettings_dump_to_file` (format by extension) / `_dump_to_string`
- `ZarrConfigFormat { Yaml, Json }`

**Python**: `StreamSettings.from_file` / `from_string` / `from_dict` and `to_file` / `to_yaml` / `to_json` / `to_dict` — routed through the C++ core (and, for dict, Python's `json`) so there's a single parse/validate/emit path with no duplicated schema logic.

## Design notes

- **Secrets are never read from config** — S3 credentials still come from the environment/IAM; the file only carries endpoint/bucket/region.
- A top-level **`version`** field gates future schema changes.
- Booleans are strict JSON-style `true`/`false` (yaml-cpp's YAML 1.1 would otherwise read a dimension named `y` as a boolean).
- On **YAML dump**, strings that would otherwise reload as a bool/int/float/null (e.g. a numeric well name like `"5"`) are emitted double-quoted, so dump→load round-trips losslessly. JSON dump is lossless by construction.
- Ownership: the loader owns the whole tree and frees it via `destroy_loaded`; the Python binding deep-copies into its own objects.
- Adds **yaml-cpp** (`>= 0.8.0`) as a vcpkg dependency.

## Tests / examples

- C++ integration `stream-from-config-file`: identical YAML vs JSON load, dump→reload round-trip (both formats), loaded settings drive a real stream, malformed configs rejected.
- Python `test_settings.py`: parametrized YAML/JSON load, round-trip through `to_yaml`/`to_json`/`to_file` and `to_dict`/`from_dict`, ambiguous-string (numeric well name) YAML round-trip, malformed→`ValueError`.
- Examples under `examples/config/` (`flat.yaml`, `flat.json`, `hcs.yaml`).

### DIFF
--- a/src/streaming/settings.io.cpp
+++ b/src/streaming/settings.io.cpp
@@ -14,7 +14,8 @@
 using json = nlohmann::json;
 
 // ---------------------------------------------------------------------------
-// YAML <-> JSON bridge (JSON is a subset of YAML, so one parse path serves both)
+// YAML <-> JSON bridge (JSON is a subset of YAML, so one parse path serves
+// both)
 // ---------------------------------------------------------------------------
 namespace {
 json
@@ -26,7 +27,8 @@ scalar_to_json(const YAML::Node& node)
     }
 
     // Strict JSON-style booleans only. yaml-cpp follows YAML 1.1, where y/yes/
-    // no/on/off are booleans -- that would turn a dimension named "y" into true.
+    // no/on/off are booleans -- that would turn a dimension named "y" into
+    // true.
     if (s == "true") {
         return true;
     }
@@ -78,44 +80,80 @@ yaml_to_json(const YAML::Node& node)
     return nullptr;
 }
 
-YAML::Node
-json_to_yaml_node(const json& doc)
+// A string needs quoting on emit iff loading it back would not yield a string
+// (mirrors scalar_to_json). Otherwise a well named "5" would round-trip to int.
+bool
+reparses_as_nonstring(const std::string& s)
 {
-    YAML::Node node;
+    if (s == "true" || s == "false") {
+        return true;
+    }
+    if (s.empty() || s == "~" || s == "null") {
+        return true;
+    }
+    const YAML::Node n(s);
+
+    if (int64_t i; YAML::convert<int64_t>::decode(n, i)) {
+        return true;
+    }
+
+    if (uint64_t u; YAML::convert<uint64_t>::decode(n, u)) {
+        return true;
+    }
+
+    if (double d; YAML::convert<double>::decode(n, d)) {
+        return true;
+    }
+
+    return false;
+}
+
+void
+emit_json(YAML::Emitter& e, const json& doc)
+{
     switch (doc.type()) {
-        case json::value_t::null:
-            node = YAML::Node(YAML::NodeType::Null);
-            break;
         case json::value_t::object:
+            e << YAML::BeginMap;
             for (const auto& [key, value] : doc.items()) {
-                node[key] = json_to_yaml_node(value);
+                e << YAML::Key << key << YAML::Value;
+                emit_json(e, value);
             }
+            e << YAML::EndMap;
             break;
         case json::value_t::array:
+            e << YAML::BeginSeq;
             for (const auto& value : doc) {
-                node.push_back(json_to_yaml_node(value));
+                emit_json(e, value);
             }
+            e << YAML::EndSeq;
             break;
-        case json::value_t::string:
-            node = doc.get<std::string>();
+        case json::value_t::string: {
+            const auto s = doc.get<std::string>();
+            if (reparses_as_nonstring(s)) {
+                e << YAML::DoubleQuoted; // one-shot: applies to the next scalar
+            }
+            e << s;
             break;
+        }
         case json::value_t::boolean:
-            node = doc.get<bool>();
+            e << doc.get<bool>();
             break;
         case json::value_t::number_integer:
-            node = doc.get<int64_t>();
+            e << doc.get<int64_t>();
             break;
         case json::value_t::number_unsigned:
-            node = doc.get<uint64_t>();
+            e << doc.get<uint64_t>();
             break;
         case json::value_t::number_float:
-            node = doc.get<double>();
+            e << doc.get<double>();
+            break;
+        case json::value_t::null:
+            e << YAML::Null;
             break;
         default:
-            node = doc.dump();
+            e << doc.dump();
             break;
     }
-    return node;
 }
 } // namespace
 
@@ -131,7 +169,7 @@ struct EnumEntry
     int value;
 };
 
-const EnumEntry kDataTypes[] = {
+constexpr EnumEntry kDataTypes[] = {
     { "uint8", ZarrDataType_uint8 },     { "uint16", ZarrDataType_uint16 },
     { "uint32", ZarrDataType_uint32 },   { "uint64", ZarrDataType_uint64 },
     { "int8", ZarrDataType_int8 },       { "int16", ZarrDataType_int16 },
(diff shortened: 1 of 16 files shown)


src/streaming/settings.io.cpp
```diff
@@ -0,0 +1,1043 @@
...
+        out.push_back(as_string(e, ctx));
+    }
+    return out;
+}
+
+const char**
+dup_string_list(const std::vector<std::string>& src)
+{
+    auto* arr = alloc_zeroed<const char*>(src.size());
+    for (size_t i = 0; i < src.size(); ++i) {
```
jeskesen, line 557: `dup_string_list` is not exception-safe. If any `dup_cstr` call throws,`arr` (the allocated pointer array) and all already-copied strings are leaked, and the function never returns — so the caller's pointer field (e.g. `plate->row_names`) is never assigned. The caller's *count* field was already set before this call, so `destroy_loaded_settings` later sees count > 0 with a null pointer and crashes indexing through it.

Suggestion: wrap the fill loop in a try/catch that frees arr[0..i-1] and then arr itself before rethrowing, or restructure load_plate to set the count only after the assignment succeeds.

src/streaming/settings.io.cpp
```diff
@@ -0,0 +1,1043 @@
...
+        }
+        std::free(s->arrays);
+    }
+
+    if (s->hcs_settings) {
+        for (size_t p = 0; p < s->hcs_settings->plate_count; ++p) {
+            auto& plate = s->hcs_settings->plates[p];
+            free_cstr(plate.path);
+            free_cstr(plate.name);
+            for (size_t i = 0; i < plate.row_count; ++i) {
```
jeskesen: If `dup_string_list` threw during `load_plate`, `plate.row_count` is > 0
but `plate.row_names` is null (the assignment never completed). This loop
then dereferences null and crashes — in the error-recovery path, which is
the worst place for a secondary failure.

Add a null guard:
  if (plate.row_names) {
      for (size_t i = 0; i < plate.row_count; ++i)
          free_cstr(plate.row_names[i]);
      std::free(...);
  }

Same fix needed for `column_names` a few lines below.

src/streaming/settings.io.cpp
```diff
@@ -0,0 +1,1043 @@
...
+// ---------------------------------------------------------------------------
+// YAML <-> JSON bridge (JSON is a subset of YAML, so one parse path serves
+// both)
+// ---------------------------------------------------------------------------
+namespace {
+json
+scalar_to_json(const YAML::Node& node)
+{
+    const auto& s = node.Scalar();
+    if (node.Tag() == "tag:yaml.org,2002:str") {
```
jeskesen: This branch is never taken. yaml-cpp 0.8.0 assigns tag `'!'` to
double-quoted and single-quoted scalars, not `'tag:yaml.org,2002:str'`
(that form only appears with explicit YAML tags like `!!str`). As written,
every quoted scalar falls through to the type-detection chain below.

The practical consequence: a double-quoted empty string (`""`) hits
`s.empty()` and returns `json(nullptr)` instead of `json("")`, breaking
any config field that uses one.

Fix: `node.Tag() == "!" || node.Tag() == "tag:yaml.org,2002:str"`

src/streaming/settings.io.cpp
```diff
@@ -0,0 +1,1043 @@
...
+            fail(ctx, "expected an integer");
+        }
+    }
+    fail(ctx, "expected an integer");
+}
+
+uint32_t
+as_u32(const json& j, const std::string& ctx)
+{
+    return static_cast<uint32_t>(as_u64(j, ctx));
```
jeskesen: `static_cast<uint32_t>` silently truncates any value above 4294967295
with no error. `array_size_px: 5000000000` becomes 705032704;
`max_threads: 4294967296` becomes 0 (all hardware threads). The
double-cast for `level` and `shuffle` (u64 → u32 → u8 at the call
sites) means `level: 256` silently becomes 0, disabling compression.

Add a range check before the cast:
    if (v > std::numeric_limits<uint32_t>::max())
        fail(ctx, "value out of range for uint32");

jeskesen (changes requested):
1. destroy_loaded_settings crashes (null dereference) when OOM hits during load_plate — src/streaming/settings.io.cpp lines 891-898 — CONFIRMED

load_plate sets plate->row_count = rows.size() before calling plate->row_names = dup_string_list(rows). If dup_string_list throws bad_alloc partway through (after alloc_zeroed succeeds but before all dup_cstr calls complete), the assignment never happens and plate->row_names stays null. The exception propagates to load_from's catch, which calls destroy_loaded_settings. That function loops for (size_t i = 0; i < plate.row_count; ++i) free_cstr(plate.row_names[i]) with no null guard on plate.row_names — indexing through a null pointer. Identical issue for column_count/column_names at lines 895–898. The comment in json_to_settings ("a throw anywhere leaves a struct safe to pass to destroy_loaded_settings") describes the intent but load_plate violates it.

Fix: null-guard the pointer before the loop, or — better — swap the ordering so the count is set only after the pointer assignment succeeds, and add a try/catch to dup_string_list that frees its partially-filled array before rethrowing.

---
2. scalar_to_json string-tag check is a dead branch; empty double-quoted YAML strings become JSON null — src/streaming/settings.io.cpp line 25 — CONFIRMED

The guard if (node.Tag() == "tag:yaml.org,2002:str") is never true for any scalar produced by normal yaml-cpp 0.8.0 parsing. yaml-cpp assigns '!' to double-quoted and single-quoted scalars and '?' to unquoted ones — tag:yaml.org,2002:str only appears if the user writes explicit YAML tags like !!str. Double-quoted empty strings ("") fall through the entire chain and hit s.empty() → return nullptr (JSON null), breaking any config field that uses them. Double-quoted numeric strings like "5" (the stated use-case for column names) become JSON integers; as_string coerces those back to strings so numeric fields survive, but only by accident.

Fix: change the guard to node.Tag() == "!" || node.Tag() == "tag:yaml.org,2002:str".

---
3. ZarrStreamSettings_load_from_file silently leaksry — src/streaming/settings.io.cpp line 734 —PLAUSIBLE

json_to_settings begins with std::memset(out, 0, sizeof(*out)). If a caller first calls ZarrStreamSettings_create_arrays(&s, n)
(which allocates with new[]) and then calls ZarrStr, path), the memset discards the new[] pointer withno free. The docs describe settings as [out] and warn against mixing lifecycle functions, but there is no runtime guard. This
differs from the calloc/delete[] mismatch (which ismemset path is unguarded and produces a silent leak.

---
4. as_u32 silently truncates values greater than UINT32_MAX — src/streaming/settings.io.cpp line 291 — CONFIRMED

as_u32 is static_cast<uint32_t>(as_u64(j, ctx)) with no range check. array_size_px: 5000000000 silently becomes 705032704; max_threads: 4294967296 silently becomes 0 (which tall hardware threads"). No downstream validationcatches the wrap-around. Affects every uint32 field: array_size_px, chunk_size_px, shard_size_chunks, max_levels, max_threads, and the double-truncated level/shuffle (uint64 → uint32tly becomes 0 — disabling compression).

src/streaming/settings.io.cpp
aliddell: Good catch — confirmed yaml-cpp 0.8.0 tags quoted scalars `"!"`, so this branch never fired and a double-quoted `""` fell through to `s.empty()` → `null`. Fixed: now `node.Tag() == "!" || node.Tag() == "tag:yaml.org,2002:str"`.

src/streaming/settings.io.cpp
aliddell: Fixed. `as_uint<T>` now range-checks against `std::numeric_limits<T>::max()` and fails instead of truncating.

src/streaming/settings.io.cpp
aliddell: Fixed both ways: `dup_string_list` now wraps the fill loop in try/catch that frees the partial array before rethrowing, and `load_plate` sets `row_count`/`column_count` only *after* the pointer assignment succeeds, so a throw never leaves count > 0 with a null pointer.

src/streaming/settings.io.cpp
aliddell: Fixed — added `if (plate.row_names)` / `if (plate.column_names)` guards around both free loops. (Combined with setting the counts after assignment in `load_plate`, the invariant now holds from both sides.)

jeskesen (approved):

jeskesen (approved):

PR #237, opened 2026-06-27
Close sealed chunk/shard file handles so reclaimed space is freed

jeskesen:
## Problem
FileHandlePool caps open handles at get_max_active_handles(), which returns
the full RLIMIT_NOFILE soft limit. Where that limit is large (e.g. 1048576),
the LRU cap is effectively never reached, so a handle opened for a
chunk/shard file stays cached in the pool for the entire lifetime of the
stream.

For a long-running stream that means every sealed chunk/shard file keeps an
open fd. If an external process deletes finished files after transferring
them elsewhere, the unlink removes the directory entry but the blocks are
not freed while the writer holds the fd (deleted-but-open inode on
Linux/XFS). Local storage fills even though the visible tree stays bounded.
Observed on a multi-day filesystem stream: thousands of deleted-but-open
chunk fds, free space falling steadily while du stayed flat.

## Fix
Close a file's pooled handle the moment its sink is finalized. A finalized
FileSink never writes its file again, so its destructor evicts the handle
via a new FileHandlePool::close(filename) (a no-op if absent or still
borrowed). The fd is released at the natural end of the file's life, so a
later unlink frees the blocks immediately. No open/close churn is added
during writing: handles are still cached and reused across a sink's writes;
only the finished file is closed at finalize.

## Validation
- Sharded [t, g, y, x] store: open fds stay bounded to the in-progress
  working set (single digits) instead of growing linearly with sealed shards.
- Unlinking sealed shards mid-stream frees disk (deleted-but-open gap gone).
- 0 open fds after close() (unchanged).
- Python suite (test_stream.py, test_dimension_transposition.py): 62 passed.

## Changes
- FileHandlePool::close(filename): close + evict an idle pooled handle.
- ~FileSink() calls it on finalize/destroy.

### DIFF
--- a/src/streaming/file.handle.cpp
+++ b/src/streaming/file.handle.cpp
@@ -109,6 +109,26 @@ zarr::FileHandlePool::return_handle(const std::string& filename)
 }
 
 void
+zarr::FileHandlePool::close(const std::string& filename)
+{
+    std::unique_lock lock(mutex_);
+
+    const auto it = cache_.find(filename);
+    if (it == cache_.end() || it->second.refcount > 0) {
+        // Not cached, or still borrowed by an in-flight write: leave it. A
+        // borrowed handle is reclaimed by the next return_handle eviction once
+        // the borrow ends.
+        return;
+    }
+
+    lru_order_.erase(it->second.lru_it);
+    cache_.erase(it); // destroys FileHandle -> close
+    cache_space_available_ = cache_.size() < max_active_handles_;
+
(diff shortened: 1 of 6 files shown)


aliddell (commented):
Looks good, simple and straightforward. Couple of things:

- could we get docstrings in the doxygen format?
- could we get a test?
- there are merge conflicts -- let me know if you want to to look at them or if you want to take them on yourself

aliddell (approved):

PR #238, opened 2026-06-30
Fix file reopen failure crashing the writer during append (regression from #237)

jeskesen:
## Summary
=Fixes a regression from #237: a failed file reopen during
`write()` crashes the writer instead of being handled as a recoverable I/O
error. Opening early as a draft to get eyes on the failing test and confirm
the intended behavior while I build out the fix.

## Background
#237 (close sealed chunk/shard fds at finalize) changed the file-handle
lifecycle: a finalized file's handle is now closed and evicted from the
FileHandlePool, so writing that file again must reopen it via
`get_handle()` -> `open()`. Before #237 the handle stayed open for the stream's
lifetime, so a reopen never happened.

## The bug
If that reopen's `open()` fails -- e.g. the file or its parent directory was
removed concurrently by an external process -- `get_handle()` throws
`std::runtime_error` (from `init_handle()`). In `FileSink::write()` the
`get_handle()` call sits **outside** the try/catch, so the throw escapes
`write()`, propagates through the worker, and surfaces as a fatal
`Failed to append data to Zarr stream: Internal error` that terminates the
writer for that array. A recoverable I/O event becomes an unrecoverable crash.

The new `file-sink-reopen-after-close` test reproduces this deterministically
(write + finalize, remove the directory, write again) and asserts the intended
contract that `write()` returns false rather than throwing. It is expected to
FAIL on current main (CI shows the throw).

### DIFF
--- a/src/streaming/file.handle.cpp
+++ b/src/streaming/file.handle.cpp
@@ -1,5 +1,6 @@
 #include "definitions.hh"
 #include "file.handle.hh"
+#include "macros.hh"
 
 void*
 init_handle(const std::string& filename, const void* flags);
@@ -63,12 +64,28 @@ zarr::FileHandlePool::get_handle(const std::string& filename)
 
     auto it = cache_.find(filename);
     if (it == cache_.end()) {
-        // not in cache, open a new handle
+        // Open the new handle BEFORE mutating cache_/lru_order_. If the open
+        // fails (e.g. the parent directory was removed concurrently), the pool
+        // state is left untouched and the failure is reported to the caller as
+        // a null handle rather than thrown. FileSink::write()/flush_() check
+        // for the null handle and return false, so a recoverable I/O error no
+        // longer escapes get_handle() as an exception that would terminate the
+        // writer. (Opening first also avoids leaving an orphan lru_order_ entry
+        // when the open throws.)
+        std::shared_ptr<FileHandle> handle;
+        try {
+            handle = std::make_shared<FileHandle>(filename);
+        } catch (const std::exception& e) {
+            LOG_ERROR(
+              "Failed to open file handle for '", filename, "': ", e.what());
+            return BorrowedHandle();
+        }
+
         lru_order_.push_front(filename);
         auto [new_it, _] =
           cache_.emplace(filename,
                          CacheEntry{
-                           .handle = std::make_shared<FileHandle>(filename),
+                           .handle = std::move(handle),
                            .lru_it = lru_order_.begin(),
                            .refcount = 0,
                          });
--- a/src/streaming/file.handle.cpp
+++ b/src/streaming/file.handle.cpp
@@ -2,6 +2,9 @@
 #include "file.handle.hh"
 #include "macros.hh"
 
+#include <chrono>
+#include <thread>
+
 void*
 init_handle(const std::string& filename, const void* flags);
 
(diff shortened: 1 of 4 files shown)


src/streaming/file.handle.cpp
```diff
@@ -63,16 +67,64 @@ zarr::FileHandlePool::get_handle(const std::string& filename)
...
+        // pool state untouched; on persistent failure return a null handle,
+        // which FileSink::write()/flush_() report as a recoverable false rather
+        // than letting an exception escape and terminate the writer.
+        constexpr int max_open_attempts = 3;
+        constexpr auto open_retry_backoff = std::chrono::milliseconds(10);
+        std::shared_ptr<FileHandle> handle;
+        bool relocked = false;
+        for (int attempt = 1;; ++attempt) {
+            try {
+                handle = std::make_shared<FileHandle>(filename);
```
aliddell, line 83: In principle another thread could do one of two things while this loop is executing:
1. Open up a different file, causing `cache_space_available_` to flip to false
2. Open up the same file, causing `make_shared()` to throw.

(2) is accounted for, but (1) is not. We could check here in the loop and `cv_.wait()` again.

jeskesen: Indeed. Though one problem with checking inside the loop: any check we do there could still go stale, since if that retry also fails, we drop the lock again for the backoff sleep before we ever get to `cache_.emplace(...)`. So a check-and-wait at the top of an iteration doesn't actually guarantee validity at insertion time — only a check made after the loop exits (right before we touch `cache_`) is guaranteed not to be invalidated by a subsequent unlock.

Suggestion: add the `cv_.wait()` once, right after the loop, gated on `relocked` (i.e., only if the lock was ever dropped). What do you think?

(EDIT: This would replace the current `if (relocked) { it = cache_.find(filename); }` block further down, around line 114-116 — not shown as a suggestion here since that's not this line.)

```cpp
        if (relocked) {
            cv_.wait(lock, [&] {
                return cache_.contains(filename) || cache_space_available_;
            });
            it = cache_.find(filename);
        }
```

aliddell (approved):
## Code Review: PR #238 — Fix file reopen failure crashing the writer during append

### Summary
Regression fix for #237, which started closing/evicting a finalized file's pooled handle. That made `get_handle()` need to reopen the file on the next write, and a failed reopen (e.g., the parent directory removed concurrently) threw `std::runtime_error` from a call site (`FileSink::write()`/`flush_()`) that sits outside any try/catch — crashing the writer thread on what should be a recoverable I/O error. The fix catches the open failure inside `get_handle()`, retries transiently (3 attempts, 10ms backoff, dropping the pool mutex during the wait), and returns a null `BorrowedHandle` on persistent failure, which both call sites already handle by returning `false`.

I compiled this branch from scratch (WSL/g++13/vcpkg) and ran it, not just read the diff:
- Both new tests (`file-sink-reopen-after-close`, `file-sink-reopen-retry-recovers`) pass.
- Pre-existing `file-sink-write` and `file-sink-close-on-destroy` tests still pass — no regression in the touched area.
- Full project builds clean; the only compiler warning is a pre-existing macro issue in `macros.hh` unrelated to this change.

### Critical Issues
None found.

### Suggestions
| # | File | Line | Suggestion | Category |
|---|------|------|------------|----------|
| 1 | `src/streaming/file.handle.cpp` | 64-66, 117-119 | The `cache_.contains(filename) \|\| cache_space_available_` predicate is duplicated between the initial `cv_.wait` and the post-retry `cv_.wait`. Minor — could factor into a small private lambda/method for maintainability. | Maintainability |
| 2 | `src/streaming/file.handle.cpp` | 116-121 | When the retry loop finds another thread already cached the same file (post-relock), that entry is used as-is without splicing it to the front of `lru_order_`, unlike the normal "found in cache" path. Harmless today since refcount increments right after, making it eviction-ineligible while borrowed, but worth a comment or a one-line fix for consistency. | Correctness (minor) |
| 3 | `src/streaming/file.handle.cpp` | 81-84 | The first open attempt (and any retry that succeeds without unlocking) still runs the blocking `FileHandle` constructor while holding `mutex_` — pre-existing pattern, now also exercised across retries. Not a regression, but a candidate for future improvement if open() latency becomes a contention concern. | Performance |

### What Looks Good
- Root cause correctly identified: the throw was escaping across an API boundary callers didn't expect to throw for this failure mode.
- Retry design opens the handle *before* mutating `cache_`/`lru_order_` so a failure leaves pool state untouched, and drops the mutex only during the backoff sleep, not during the open syscalls.
- The follow-up commit (`29408fd`) closing the "cache filled up while I was unlocked" race by re-waiting on `cache_space_available_` / re-checking for a concurrent insert after relocking shows real attention to the concurrency implications of dropping and reacquiring the lock.
- Test coverage is well-targeted and deterministic (directory removal/recreation, not timing-probabilistic): one test proves the give-up path returns `false` instead of throwing, the companion test proves the retry actually recovers a transient failure.
- Commit history isolates one concern per commit (failing test → fix → retry → race fix) with messages that explain *why*.

### Verdict
Approve.

PR #239, opened 2026-07-13
Add ZARR_MAX_THREADS environment variable override

jeskesen:
## Use case

I'm calling acquire-zarr from ome-writers, where I don't have access to
the full settings API — I need a way to control the thread pool size
without plumbing a new parameter through. An environment variable lets
me set this from the Python side without touching the API surface
ome-writers exposes.

## Plan

- `zarr::resolve_max_threads` (src/streaming/zarr.common.hh/cpp) reads
  `ZARR_MAX_THREADS` and applies it only when `max_threads` wasn't
  explicitly set (0). Invalid values (non-numeric, <= 0, overflow) log a
  warning and fall back to auto-detection.
- Wired into `ZarrStream_s::start_thread_pool_`, so it's honored by every
  entry point (C API, C++, the JSON/YAML config loader, and Python
  bindings) since they all funnel through the same constructor.
- Still to do in this PR: `PyZarrStreamSettings.max_threads_` currently
  defaults to `hardware_concurrency()` instead of `0`, so an unset
  `max_threads` looks "explicit" by the time it reaches the C++ layer and
  the env var never takes effect for a typical Python caller. Fixing
  that default plus adding a Python-level test is the remaining work
  before this is ready for review.

Opening as draft to get CI running early; will mark ready once the
Python-side fix and tests land.

## Test plan
- [x] New unit test `resolve-max-threads` covering precedence and
      invalid-value handling
- [x] `create-stream` integration case exercising the real thread pool
      wiring end-to-end
- [x] Python-level test for the env var via the Python bindings

### DIFF
--- a/src/streaming/zarr.stream.cpp
+++ b/src/streaming/zarr.stream.cpp
@@ -1428,6 +1428,7 @@ ZarrStream_s::commit_settings_(const ZarrStreamSettings* settings)
 void
 ZarrStream_s::start_thread_pool_(uint32_t max_threads)
 {
+    max_threads = zarr::resolve_max_threads(max_threads);
     max_threads =
       max_threads == 0 ? std::thread::hardware_concurrency() : max_threads;
     if (max_threads == 0) {
@@ -1807,4 +1808,10 @@ finalize_stream(ZarrStream* stream)
     }
 
     return true;
+}
+
+uint32_t
+stream_thread_count(ZarrStream* stream)
+{
+    return stream->thread_pool_->n_threads();
 }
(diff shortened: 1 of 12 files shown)


src/streaming/zarr.stream.cpp
```diff
@@ -1808,3 +1809,9 @@ finalize_stream(ZarrStream* stream)
 
     return true;
 }
+
+uint32_t
+stream_thread_count(ZarrStream* stream)
```
aliddell: `stream` could be `const` here

aliddell (approved):
LGTM. One small change you could make or could be picked up later.

PR #242, opened 2026-08-11
Make CITATION.cff valid CFF 1.2.0

aliddell:
`CITATION.cff` does not validate against CFF 1.2.0. This fixes it.

## The violations

Confirmed individually with `cffconvert --validate`, isolating one field at a time:

| Field | Error |
|---|---|
| `message` | `'message' is a required property` — CFF 1.2.0 requires `authors`, `cff-version`, `message`, `title` |
| `license` | `'apache-2.0' is not one of [...]` — SPDX identifiers are case-sensitive; the enum has `Apache-2.0` |
| `orcid` | `'0000-0001-6236-9282' does not match 'https://orcid\.org/[0-9]{4}-[0-9]{4}-[0-9]{4}-[0-9]{3}[0-9X]{1}'` |

`message` has been absent since the file was added in #56, so the file has never been valid.

After this change:

```
$ cffconvert --validate -i CITATION.cff
Citation metadata are valid according to schema version 1.2.0.

$ cffconvert -f apalike -i CITATION.cff
Liddell A., Eskesen J., Clack N. (2026). acquire-zarr: Streaming directly
to Zarr on the file system or cloud (version 0.9.0). DOI: 10.5281/zenodo.14828040
URL: https://github.com/acquire-project/acquire-zarr
```

## How it was found

The Zenodo GitHub integration was enabled for the v0.9.0 release and rejected it:

```json
{"error_id": "464593eb36784fabacee037683210fee", "errors": "Citation metadata load failed"}
```

GitHub delivered the `published` event and Zenodo returned 202, then failed while parsing this file, before creating any deposit. Since `message` was never present, that integration could never have worked for any release — which is likely why the only Zenodo deposit was a hand-uploaded `acquire-zarr-0.1.0.tar.gz` from 2025-02-06 rather than anything the webhook produced.

## Why it still matters, given the webhook is staying off

The Zenodo GitHub integration is deliberately **not** the archiving route, and has been turned back off. It can only create a new record under a new concept DOI; it cannot add versions to the existing concept record `10.5281/zenodo.14828040`, which is what the README badge and any existing citations point at. Releases are instead archived as new versions of that record through the Zenodo API, keeping a single lineage.

So this fix is not about the webhook. A valid `CITATION.cff` is needed for:

- GitHub's "Cite this repository" widget, which parses this file
- Any CFF consumer — `cffconvert`, Zotero, citation.js, Software Heritage
- The planned `release.yml` archiving step, which derives Zenodo metadata from this file via `cffconvert -f zenodo`

## Also here

- Bumps `version`/`date-released` to 0.9.0 and adds `repository-code`
- Updates author affiliations to Chan Zuckerberg Biohub

## Follow-up

`doi:` still points at the concept DOI. Once 0.9.0 is archived, a follow-up sets `doi:` to the 0.9.0 version DOI and moves the concept DOI into `identifiers:`, as recommended for CFF. The README badge needs no change — it already uses the concept DOI, which resolves to the newest version.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -11,7 +11,11 @@ authors:
   orcid: https://orcid.org/0000-0001-6236-9282
 cff-version: 1.2.0
 date-released: '2026-08-11'
-doi: 10.5281/zenodo.14828040
+doi: 10.5281/zenodo.21892791
+identifiers:
+- type: doi
+  value: 10.5281/zenodo.14828040
+  description: The concept DOI of the work
 license: Apache-2.0
 message: If you use this software, please cite it using the metadata from this file.
 repository-code: https://github.com/acquire-project/acquire-zarr
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -8,12 +8,13 @@ authors:
 - affiliation: Chan Zuckerberg Initiative (United States)
   family-names: Clack
   given-names: Nathan
-  orcid: 0000-0001-6236-9282
+  orcid: https://orcid.org/0000-0001-6236-9282
 cff-version: 1.2.0
-date-released: '2026-05-29'
+date-released: '2026-08-11'
 doi: 10.5281/zenodo.14828040
-license:
-- apache-2.0
+license: Apache-2.0
+message: If you use this software, please cite it using the metadata from this file.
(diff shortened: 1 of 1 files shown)


nclack (approved):
lgtm!

PR #244, opened 2026-08-18
Emit RFC-5 coordinate systems and transformations for OME-Zarr 0.6

aliddell:
Second slice of the OME-NGFF expansion: emit RFC-5 coordinate systems and
transformations when the stream selects OME-Zarr 0.6.

**Stacked on #240** — that PR adds the `ome_version` selector this one gates on.

## What changes

`MultiscaleArray::make_multiscales_metadata_()` restructures the existing
per-level `scale` transform into the RFC-5 shape when `ome_version == 0_6`:

- a `coordinateSystems` entry named `"intrinsic"`, built from the visible axes
- per-dataset `scale` transforms that name their `input` (`{"path": "<level>"}`)
  and `output` (`{"name": "intrinsic"}`)
- no top-level `axes` in 0.6, which `coordinateSystems` supersedes

0.5 output is unchanged. No new user-facing API — richer transforms
(translation/affine/rotation/sequence, named systems) need one and come later.

Design notes and the resolved open questions are in
`docs/design/rfc5-coordinate-transformations.md`.

## Notes

RFC-5 is still under review, so 0.6 stays opt-in and is pinned to 0.6.dev3.

## Testing

- `tests/integration/stream-omero-and-ome-version.cpp` asserts the 0.6 shape and
  keeps the 0.5 assertions as a regression guard
- `python/tests/test_stream.py::test_ome_version_selector`


nclack:
I worry about the approach on this one. Ideally, we wouldn't need to explicitly model every ngff feature. 

For example, if the caller can just hand us a json string that we put in the right place (like an "omero" key), then we don't need to model hardly anything here. That way we don't need to change the code as much as the standard evolves.

Issue #247, opened 2026-08-24
Consolidate path/key joining into a join_key helper


aliddell:
There are ~11 sites that build a storage path or Zarr key with `a + "/" + b`. They fall into two families that look identical but have opposite requirements, and the difference is currently handled ad hoc (or not at all).

### Family A — Zarr keys (relative; leading slash must be stripped)

- `zarr.stream.cpp:340`, `:420` — join, then `regularize_key`. Correct.
- `zarr.stream.cpp:1348`, `:1385` — the same idiom, split across two statements, duplicated verbatim.
- `zarr.stream.cpp:1202` — joins without regularizing. Works only because `validate_array_settings` regularizes downstream.
- `acquire.zarr.cpp:54` — joins without regularizing, and the result is inserted into a `unique_paths` dedup set. An empty row/column/field segment yields a key that does not match the path actually used, so this is a dedup miss rather than a cosmetic issue.

### Family B — store paths (may be an absolute filesystem path)

- `zarr.stream.cpp:1572`, `array.base.cpp:61`/`:75`, `sink.cpp:78`

These are hand-guarded with `if (!segment.empty())` in two places, and were previously unguarded in a third (fixed in the aws-crt-cpp migration: an empty root-group segment produced `store//zarr.json`, which MinIO rejects as an invalid object name. It had been invisible because the filesystem collapses duplicate separators and minio-cpp normalized the signed URI path).

### Why the obvious fix is wrong

Wrapping every join in `zarr::regularize_key` is tempting — it already strips leading/trailing slashes *and* collapses interior ones — but it would break Family B. `store_path_ = zarr::trim(settings->store_path)` only strips whitespace, so a leading `/` survives, and it has to: the same string is passed to `make_file_sink`. Regularizing it would turn `/mnt/data/x.zarr` into `mnt/data/x.zarr` and silently write to a relative path.

### Proposal

Add to `zarr.common`:

```cpp
/// Join non-empty segments with a single '/'. The first segment is kept verbatim
/// so absolute filesystem paths keep their leading slash; interior segments are
/// trimmed of surrounding slashes.
std::string join_key(std::initializer_list<std::string_view> segments);
```

- Family B becomes `join_key({store_path_, relative_path, metadata_key})`.
- Family A becomes `join_key({parent_path, key})`, with `regularize_key` retained only at the user-input boundary, where it already is (`regularize_key(settings->output_key)`).

[...]

PR #248, opened 2026-08-24
Replace minio-cpp with aws-crt-cpp for S3 storage

aliddell:
Supersedes #246, which failed every CI check.

## Why replace rather than patch

Dependabot #246 broke the build because minio-cpp rewrote its API to return `tl::expected<T, Error>` from every client call. That was the trigger; the reasons to leave are structural:

- **We tracked an unreleased branch.** In that single bump the project swapped its HTTP stack (curlpp → cpp-httplib) and vendored prebuilt `libs3rdma` binaries for amd64 and arm64.
- **It was not consumable from vcpkg.** `CMakeLists.txt` carried `#find_package(miniocpp CONFIG REQUIRED) # TODO: uncomment when minio-cpp pushes a new release to vcpkg`, so we held a git submodule *and* a `vcpkg.json` entry, `add_subdirectory`'d the submodule, and `install(TARGETS miniocpp)`'d it into our own export set. That is why `examples/vcpkg.json` had to vendor curlpp/inih/openssl/pugixml/zlib.
- **A known, unfixed crash.** It called `curl_global_init`/`curl_global_cleanup` per request, racing live transfers and segfaulting the S3 tests. A global-init-hold workaround was tried and was insufficient.

aws-crt-cpp is a real vcpkg package, so the submodule is gone. Its client is thread-safe, so the connection pool is gone too. Multipart upload, part numbering, parallel part transfer, ETag collection and retries move into the CRT.

## What changed

- **`S3ConnectionPool` deleted.** One client now serves every writer thread, instead of `hardware_concurrency()` connections sized independently of the thread pool (`max_threads=64` on an 8-core box meant 56 workers blocking in `get_connection()`). This also removes four unchecked `get_connection()` null-derefs on teardown.
- **The hand-rolled multipart state machine and the 5 MiB per-sink `part_buffer_` are deleted.** `S3Sink` now opens one `CreateWithAsyncWrites` PUT per object.
- **Bucket addressing is now explicit:** virtual-host style for `*.amazonaws.com`, path style otherwise, which is what most S3-compatible servers require. The CRT never rewrites the request path, so this could no longer be left implicit. A bucket whose name is not a legal DNS label — because it contains a dot, an underscore or an upper-case letter — falls back to path style even on AWS, since it cannot be prepended to the host and a dot would add a label the `*.s3.<region>.amazonaws.com` wildcard certificate cannot match. That is the fallback boto3 and aws-sdk-cpp apply, and the workaround AWS documents.
- **An unset `region` is signed as `us-east-1`** rather than omitted. SigV4 always needs one. Endpoints that validate the signing region — MinIO with a configured site region among them — require it to be set explicitly to match the server. The public C and Python API is unchanged.
- **Scheme and host are matched case-insensitively.** `HTTPS://minio:9000` and `https://S3.US-WEST-2.AMAZONAWS.COM` resolve the same as their lower-case forms; the latter would otherwise have slipped past the `.amazonaws.com` check and picked path style against real AWS.
- **The bucket is checked once at stream creation** instead of on every sink creation — `make_s3_sink` was issuing a `HeadBucket` per chunk, shard and metadata object.
- ⚠️ **The endpoint must now begin with `http://` or `https://`.** minio-cpp inferred TLS from an `https` prefix and otherwise fell back to plaintext, so a bare `host:port` silently meant HTTP. Neither default survives the move: plaintext silently downgrades a server that expected TLS, and assuming TLS fails the handshake against a plain-HTTP MinIO. A scheme-less endpoint is now rejected at settings validation with an actionable message. **This breaks configurations that relied on the old default**, including the scheme-less form this README used. Settings validation and the client share one parse (`zarr::is_valid_s3_endpoint`, declared beside `s3_request_target()`) so they cannot disagree about what is acceptable, and so the message is written once; `endpoint-request-target` covers both, and is deliberately not named `s3-*` so it does not inherit the label that requires MinIO.
- **`S3Sink` holds its append slot through an RAII claim.** Only one thread may append to a CRT async-write upload at a time, and that flag was originally set and cleared by hand around calls that release the lock — so an exception out of `create_upload` or `append` left it latched, and the flush at close then waited on a condition that could never become true, deadlocking instead of reporting the error. The claim now releases on every path, retaking the lock first if an append threw, and marks the sink failed unless the append is acknowledged: otherwise bytes counted as appended but never accepted would let a later flush report success on a truncated object. `flush_` holds the claim for the whole of finalization, since a writer arriving mid-flush could otherwise issue a second concurrent `Write`, which the CRT forbids. Writes that arrive once finalization has begun are refused rather than staged, as they can no longer reach the object.
- **Staged bytes are reported in memory usage.** Out-of-order writers can leave most of a shard sitting in `S3Sink`'s pending map — a second copy of its compressed bytes — where `Array::memory_usage()` could not see it. `Sink` gains an opt-in `memory_usage()` that `S3Sink` answers from an atomic; `Shard` and `Array` aggregate it with `try_to_lock` and report zero under contention, since this feeds an estimate and must not block a writer. Bytes already handed to an in-flight append are excluded: one batch at most, and counting them would leak the tally when an append throws. `ZarrStreamSettings_estimate_max_memory_usage` works from settings alone and still does not model this.
- **`Content-Length` is parsed with `from_chars`.** The `HeadObject` headers callback runs on a CRT thread invoked from C, where `stoull` would have thrown through C frames on a value from a non-conforming gateway instead of returning the documented `nullopt`. The callback now neither allocates nor logs, and an unparseable length is left unset exactly as an absent one is.

## Fixes a pre-existing data race

`Shard::write_chunk` claims each chunk's offset under `mutex_` but calls the sink *outside* it, and `Array::dispatch_chunk_job_` makes every chunk its own thread-pool job. So byte ranges reached `S3Sink` concurrently and out of order while it mutated `nbytes_buffered_`/`nbytes_flushed_`/`part_buffer_` with no synchronization. `FileSink` was unaffected because its write is positional.

Since the CRT upload is strictly append-only, the sink now stages fragments and appends them in order, which fixes the race as a side effect. New `s3-sink-write-out-of-order` writes a large object back-to-front from several threads and requires the bytes back byte-exact.

## Performance

Measured against the previous release (0.9.0, minio-cpp) on the same machine, same
interpreter and same server. Baseline is the PyPI `acquire_zarr==0.9.0` cp313 wheel;
candidate is this branch built locally as a cp313 wheel. 128 frames of 1024x1024
`uint16` (256 MiB) per repetition, y/x chunks of 256, 2x2 shards, `max_threads=8`,
three A/B/A/B interleaved repetitions. Timing spans stream creation through `close()`.

**Uncompressed**

| | 0.9.0 (minio-cpp) | this branch (aws-crt) |
|---|---|---|
| MB/s (3 reps) | 32.3 / 34.4 / 33.4 | 81.0 / 92.6 / 89.3 |
| wall clock | 7.93 / 7.45 / 7.66 s | 3.16 / 2.76 / 2.87 s |
| peak process RSS | 852 MB | 898-914 MB |

**Blosc-LZ4, level 1**

| | 0.9.0 (minio-cpp) | this branch (aws-crt) |
|---|---|---|
| MB/s (3 reps) | 60.5 / 55.1 / 52.1 | 118.6 / 150.6 / 131.9 |
| peak process RSS | 852-880 MB | 878-882 MB |

Median **2.7x faster uncompressed and 2.4x compressed**, for roughly 6% more resident
memory. RSS is whole-process and dominated by the chunk buffers held until close, so
that delta is a small increment on a large shared baseline; it is consistent with the
CRT's own connection buffers plus the sink staging this branch now reports through
`ZarrStream_get_current_memory_usage`.

### The baseline cannot be measured in steady state

Both configurations above deliberately set the append-dimension chunk size larger than
the frame count, so every shard is flushed at `close()` and no rollover happens
mid-stream. That is the only regime 0.9.0 survives: it segfaults as soon as a chunk
layer completes, reproducibly at exactly `frames == chunk_size_px` of the append
dimension, at both 512x512 and 2048x2048 frames, on the PyPI wheel and on a local
pre-migration build. The last thing the debug log shows is `rollover_`, then
`create_multipart_object` and `upload_multipart_object_part` for the four shards of the
completed layer. The same shapes written to the filesystem complete fine, as does this
branch. This is consistent with the per-request `curl_global_init`/`curl_global_cleanup`
crash described above, though I did not confirm that root cause directly.

For reference, this branch in steady state -- 256 frames, append chunk of 32, so eight
mid-stream rollovers -- sustains **169.0 and 141.4 MB/s** at 848-868 MB RSS, i.e. faster
than the flush-at-close case, because flushes overlap with appends instead of landing
all at once. There is no 0.9.0 figure to compare it against.

### Environment

- **Host:** Windows 11 Enterprise 10.0.26100, Intel Xeon W-2235 @ 3.80 GHz (6 cores /
  12 threads), 31.6 GB RAM.
- **Server:** MinIO `RELEASE.2025-09-07T16-13-09Z`, single node, `minio server /tmp/minio`,
  running under WSL2 (Ubuntu 24.04) with its data directory on the WSL ext4 volume.
- **Path:** the benchmark process runs on the Windows host and reaches the server over
  `http://localhost:9000`, so traffic crosses the WSL2 loopback forwarder. That penalty
  applies equally to both builds, but it means the absolute MB/s understate what a native
  or same-kernel deployment would show. These are relative figures, not a throughput spec.
- **Runtime:** CPython 3.13.2, numpy 2.5.2, both wheels `cp313-cp313-win_amd64`.
- Local MinIO also means this measures client-side cost -- request assembly, staging,
  threading -- rather than WAN behaviour, which is where the CRT's parallel part uploads
  would matter more.

## Version pins

The C++ S3 wrapper (`aws/crt/s3/S3.h`) only exists from aws-crt-cpp **v0.42.0**, which neither of our pins could reach, so both move to `2026.07.29` (aws-crt-cpp 0.43.0 / aws-c-s3 0.13.2):

| Pin | Before | After |
|---|---|---|
| `vcpkg-configuration.json` baseline | `f7423ee1…` (2024-06-14, crt-cpp 0.26.4) | `9e593bb1…` (crt-cpp 0.43.0) |
| CI vcpkg tag (10 sites) | `2025.03.19` (crt-cpp 0.29.7) | `2026.07.29` |

This re-resolves every other dependency and invalidates the CI vcpkg caches once. Notably `yaml-cpp` 0.8.0 → 0.9.0 and `nlohmann-json` → 3.12.0.

⚠️ **aws-crt-cpp 0.43.5 flips the meta-request callbacks from `int`/`AWS_OP_SUCCESS` to `bool`, where `0` changes meaning from "continue" to "abort"** — a silent-truncation trap on a future baseline bump. `s3.client.cpp` `static_assert`s the convention it compiled against so that fails loudly instead.

## Incidental fixes the migration forced

- `logger.hh` had **no `#pragma once`**, so it could only be included once per TU. Every other header in the tree has the guard.
- `zarr.stream.cpp` built the root group's metadata key by joining an empty segment, giving `store//zarr.json`. The filesystem collapsed the duplicate separator and minio-cpp normalized the signed URI path, so it was invisible; MinIO rejects it. Consolidating the ~11 join sites is tracked in #247.
- `s3.client.cpp` defines `NOMINMAX` ahead of every include, replacing the `#undef min` the old `s3.sink.cpp` needed for the same reason. Ahead of *every* include, not just the CRT ones, since any header may reach `windows.h` first.
- `zarr::to_lower` joins `trim` in `zarr.common`, replacing one of the three places the same `std::transform`/`std::tolower` body was inlined. The `Content-Length` header callback keeps its own comparison: it runs on a CRT thread and is deliberately allocation-free.
- `stream-append-nullptr` does S3 work but was never `s3`-labelled (the label is derived from the target name), and the `test_s3` job's Python step never set the AWS credentials, so `test_stream_data_to_s3` silently skipped. Both fixed.
- `VCPKG_FORCE_DOWNLOADED_BINARIES` has to be repeated in `CIBW_ENVIRONMENT`. cibuildwheel does not forward host env vars into the manylinux container, where the `cmake<4.0.0` build pin resolves to 3.31 — exactly the version that cannot run the new portfiles. The `Dockerfile` needed it as well, since ubuntu 24.04 ships CMake 3.28.

## Pre-existing races fixed here, not migration fallout

`ZarrStream_get_current_memory_usage` is a public entry point with no documented thread restriction, and the project's own `test_stream.py` polls it while appending. Three of these four are bugs on `main`, unrelated to the S3 migration; they are fixed here because tracing the first one surfaced them.

- **`close_sinks_()` cleared `shards_` without `shards_mutex_`.** This is the one race the migration introduced, since `memory_usage()` did not touch `shards_` before. It did not merely skip a lock — it defeated the `try_lock` on the reader side, which *succeeded* because no writer held the mutex, and then walked a vector the frame thread was freeing. `rollover_()` opens the window once per append-shard boundary. `close_sinks_()` now swaps the vector out under the lock and lets the shard destructors' flush I/O run outside it, so a poll during a flush does not silently drop the shard term.
- **The `chunks_` loop in the same function had the same defect.** The slot was read unlocked while `dispatch_chunk_job_` moved it out under `chunk_mutexes_[i]`, so the `Chunk` could be destroyed between `if (chunk)` and `chunk->size_bytes()`. Each slot is now read under its own try-lock. The old comment justified the wrong thing: `size_bytes()` is a const scalar, but the slot pointer's lifetime is the hazard.
- **A missed wakeup could hang `close_()` forever.** Workers retired outstanding writes with a bare `fetch_sub` and notify, while `close_()` evaluates its predicate under `write_counter_mutex_`. A notify landing between that evaluation and the wait's atomic release-and-block is lost; if it was the last write, `close_()` never wakes. Both sites now go through `finish_write_()`, which decrements under the mutex. The notify on the increment side is dropped — the only waiter's predicate is `== 0`.
- **`get_memory_usage()` read state being torn down.** It dereferenced `output->array` while `finalize_stream` moved it out, and read `output->frame_buffer.size()` while the appending thread resized it. The array slots are now guarded by `arrays_mutex_`, try-locked by the reader so an estimate never blocks a close, and the buffer size is published through an atomic.

`ArrayBase::memory_usage()` now documents the contract the implementations keep: callable from any thread, best effort, never blocks a writer. Every lock on that path is a try-lock, so a busy buffer is omitted from the estimate rather than waited for.

## Tests

Six tests reached past the client into minio types for verification and teardown, four of them carrying near-identical copies of the same ~90-line helper block. They now share `tests/s3-test-helper.hh`. `miniocpp::miniocpp` was linked into all 51 test targets, not just the S3 ones; that's removed rather than replaced, since only the helper needs a client.

`endpoint-request-target` needs no server, and covers the addressing rules directly: the dotted, upper-case, underscore and leading-hyphen buckets that must fall back to path style, the DNS-label bucket that must not, the case-folded endpoints, and `zarr::is_valid_s3_endpoint` itself — so settings validation is exercised on the same cases as the client.

C++ and Python suites pass locally against MinIO, including the `s3`-labelled tests.

## Not covered

- **Real-AWS virtual-host addressing has no automated coverage** — CI only exercises MinIO path-style. The emitted request target is asserted in `endpoint-request-target`, including the dotted-bucket fallback, but nothing sends a request to AWS. Verified by inspection only.
- `aws-c-s3` declares `"supports": "!(windows & arm) & !uwp"`, so **Windows-on-ARM is no longer supported**, which minio-cpp presumably was.


src/streaming/array.cpp
```diff
@@ -147,6 +147,18 @@ zarr::Array::memory_usage() const noexcept
         }
     }
 
+    // An S3 shard stages fragments that arrived out of order, which can reach
+    // most of the shard. Skip the tally rather than wait for it: this is an
+    // estimate, and callers must not block on the write path.
+    std::unique_lock shards_lock(shards_mutex_, std::try_to_lock);
```
jeskesen: **This new reader races `close_sinks_()`, which clears `shards_` without holding `shards_mutex_`.**

`make_shards_()` (line 467) and `finalize_shards_()` (line 968) both take `shards_mutex_`, and `memory_usage()` here takes it under `try_to_lock` before iterating `shards_`. But `close_sinks_()` (line 982) is called from `rollover_()` on the frame-processing thread and clears `shards_` with no lock at all.

So a caller invoking the public `ZarrStream_get_current_memory_usage()` from another thread during a mid-stream rollover can iterate a `std::vector<std::shared_ptr<Shard>>` while it is being cleared: dangling iterators / use-after-free. Before this PR `memory_usage()` never touched `shards_`, so the race is new here. Suggest taking `shards_mutex_` in `close_sinks_()`.

src/streaming/s3.sink.cpp
```diff
@@ -1,53 +1,130 @@
...
+    bool retval = true;
+
+    while (retval && !staged_.empty() &&
+           (upload_ || staged_.size() >= min_part_size_)) {
+        if (!upload_) {
+            lock.unlock();
+            auto upload = client_->create_upload(bucket_name_, object_key_);
+            lock.lock();
+
+            if (!upload) {
```
jeskesen: **A transient failure to *start* the upload permanently poisons the sink, defeating the caller's retry loop.**

When `create_upload()` returns nullptr, `drain_` sets `retval = false` and never acknowledges the `AppendClaim`, whose destructor sets `failed_ = true`. But no bytes were handed to the upload at this point, so the state is perfectly retryable: the staged bytes are still intact and in order.

Because every subsequent `write()` short-circuits on `failed_`, neither the 3-attempt retry loop in `array.cpp:709` nor the 2-attempt `execute_job_with_retry` can recover: one blip in the CreateMultipartUpload request fails the whole stream. Consider leaving `failed_` unset on this specific path.

src/streaming/s3.client.cpp
```diff
@@ -0,0 +1,755 @@
...
+    parsed.authority = std::string(rest);
+
+    // AWS requires virtual-host addressing; MinIO and most other S3-compatible
+    // servers require path style. Decide from the endpoint rather than adding a
+    // knob callers would have no way to know they must set.
+    std::string_view host = parsed.authority;
+    if (const auto colon = host.rfind(':'); colon != std::string_view::npos) {
+        host = host.substr(0, colon);
+    }
+    parsed.virtual_host = host.ends_with(".amazonaws.com");
```
jeskesen: **Virtual-host addressing is forced for all `*.amazonaws.com` endpoints, which breaks dotted bucket names over HTTPS.**

For bucket `my.data`, `make_target` produces `https://my.data.s3.amazonaws.com`. AWS's wildcard certificate `*.s3.amazonaws.com` does not match a name with an extra label, so TLS hostname verification fails for any legal bucket containing a dot. minio-cpp's path-style requests worked here.

There is no validation rejecting dotted bucket names either (`validate_s3_settings` only checks length/charset), so the user gets an opaque handshake error. Worth at least falling back to path style when the bucket name contains a `.`.

src/streaming/s3.client.cpp
```diff
@@ -0,0 +1,755 @@
...
+Endpoint
+parse_endpoint(const std::string& endpoint)
+{
+    Endpoint parsed;
+
+    // Neither default is safe to guess: TLS would fail the handshake against a
+    // plaintext server, and plaintext would silently downgrade one that
+    // expected TLS. Make the caller say which they meant.
+    std::string_view rest(endpoint);
+    EXPECT(rest.starts_with("http://") || rest.starts_with("https://"),
```
jeskesen: **Scheme and host matching are case-sensitive, but URL schemes and hostnames are not.**

This `EXPECT` rejects the perfectly legal `HTTPS://minio:9000`. Relatedly, `host.ends_with(".amazonaws.com")` below misses `https://S3.AMAZONAWS.COM` and silently selects path-style addressing against real AWS, which rejects path style for buckets created after Sept 2020.

Suggest lowercasing the scheme and host before comparing.

src/streaming/s3.sink.cpp
```diff
@@ -57,148 +134,133 @@ zarr::S3Sink::write(size_t offset, ConstByteSpan data)
...
-
-        retval = true;
-
-        nbytes_flushed_ = nbytes_buffered_;
-        nbytes_buffered_ = 0;
-    } catch (const std::exception& exc) {
-        LOG_ERROR("Error: ", exc.what());
+    // Shard::write_chunk replays the same offset when a write is retried, so a
+    // fragment we are still holding is a no-op rather than an error
+    if (const auto it = pending_.find(offset); it != pending_.end()) {
```
jeskesen: **Retry de-duplication only inspects `pending_`, not `staged_`.** (Latent today.)

Once `coalesce_()` has moved a fragment into `staged_`, a replayed `write()` at that same offset falls into the `else` branch and is re-inserted into `pending_` at an offset strictly below `nbytes_appended_ + staged_.size()`, which `coalesce_()` can never match. The fragment sticks forever and `flush_` fails with "is missing the bytes at offset ...".

This is unreachable right now only because every path that returns `false` after staging also sets `failed_`. It becomes a silent shard-loss bug the moment a non-poisoning failure path is added, e.g. if the `create_upload` finding is fixed as suggested. Guard by also rejecting/ignoring `offset < nbytes_appended_ + staged_.size()`.

CMakeLists.txt
```diff
@@ -81,12 +80,6 @@ install(EXPORT acquire-zarrTargets
...
 
-install(TARGETS miniocpp
-        EXPORT acquire-zarrTargets
-        LIBRARY DESTINATION lib
-        ARCHIVE DESTINATION lib
-)
-
 install(DIRECTORY examples/
         DESTINATION examples
         FILES_MATCHING
```
jeskesen: **The new `examples/vcpkg-configuration.json` is not installed.**

This `FILES_MATCHING` block lists `vcpkg.json` but not `vcpkg-configuration.json`, so the installed examples tree gets the new `aws-crt-cpp` dependency without the `9e593bb1...` baseline that guarantees >= 0.42 (the first version with `aws/crt/s3/S3.h`). A consumer building the installed examples against an older default vcpkg baseline resolves a different CRT than the library was built with.

Suggest adding a `PATTERN "vcpkg-configuration.json"`.

src/streaming/s3.client.cpp
```diff
@@ -0,0 +1,755 @@
...
+}
+
+zarr::S3RequestTarget
+make_target(const Endpoint& endpoint,
+            std::string_view bucket_name,
+            std::string_view object_name)
+{
+    zarr::S3RequestTarget target;
+
+    if (endpoint.virtual_host) {
```
nclack, line 125: Looks like this might be wrong for buckets with dots in the name. Given something like:

```
endpoint:    https://s3.us-west-2.amazonaws.com
bucket_name: microscopy.data                                                                             region:      us-west-2
```

This will make: `https://microscopy.data.s3.us-west-2.amazonaws.com`
where it should make: `https://s3.us-west-2.amazonaws.com/microscopy.data`

src/streaming/s3.client.cpp
```diff
@@ -0,0 +1,755 @@
...
+    }
+};
+
+zarr::S3Client::S3Client(const S3Settings& settings)
+  : impl_(std::make_unique<Impl>())
+{
+    impl_->endpoint = parse_endpoint(settings.endpoint);
+
+    // SigV4 always needs a region. S3-compatible servers ignore it, and callers
+    // targeting real AWS were already obliged to supply the right one.
```
nclack, line 334: Some S3-compatible servers, including MinIO configured with a site region, validate the SigV4 region

```suggestion
    // Use us-east-1 when no region is configured; endpoints that
    // validate the signing region require it to be set explicitly 
```

The same claim appears in the header and user-facing docs, so those should be reviewed too.

python/acquire_zarr/__init__.pyi
nclack, line 433: would this 's3://' url still be accepted by this pr? Might need to update examples like this one.

nclack (changes requested):
Found very minor stuff. There's one issue with buckets with dots in the name. Otherwise I think it's mostly comments that need to get cleaned up. 

I didn't note a specific example, but it's worth checking that MinIO is only mentioned in comments that are specific to testing.

src/streaming/array.cpp
jeskesen, line 173: `close_sinks_()` (line 982) clears `shards_` without taking `shards_mutex_`, so this loop can be walking the vector while it's being cleared. `rollover_()` calls it on the frame thread, and `ZarrStream_get_current_memory_usage()` is public, so a caller on another thread gets a use-after-free.

New as of this PR, since `memory_usage()` didn't touch `shards_` before. Can `close_sinks_()` just take the mutex?

src/streaming/s3.client.cpp
jeskesen: Schemes and hostnames are case-insensitive, but these comparisons aren't. `HTTPS://minio:9000` gets rejected outright, and `S3.AMAZONAWS.COM` slips past the `ends_with` below and picks path style against real AWS — which rejects it for buckets created after Sept 2020. Worth lowercasing first.

jeskesen (changes requested):
One possible concurrency snag may be worth fixing.

src/streaming/s3.client.cpp
aliddell: Good catch, fixed. A new `is_virtual_host_addressable()` applies botocore's `check_dns_name` rule so that a dot, underscore, or uppercase letter falls back to path style even on AWS (used by boto3 and aws-sdk-cpp).

src/streaming/s3.client.cpp
aliddell: Took your suggested wording and propagated it to all five sites carrying the "S3-compatible servers ignore it" claim (s3.client.cpp, s3.client.hh, README, .pyi, CHANGELOG). Also documented `bucket_name` and region in `zarr.types.h`.

python/acquire_zarr/__init__.pyi
aliddell: Fixed.

src/streaming/array.cpp
aliddell: Confirmed. Fixed by swapping under the lock and letting the shard destructors' flush I/O run outside it. Verified no deadlock: `shards_mutex_` is a strict outermost lock, neither caller holds it, and `finalize_shards_()` already holds it across the same I/O.

src/streaming/s3.client.cpp
aliddell: Consolidated into `try_parse_endpoint`, with `is_valid_s3_endpoint()` and `parse_endpoint()` as wrappers. Added a shared `zarr::to_lower`.

jeskesen (approved):
Nice.  LGTM

jeskesen (approved):

nclack (approved):
lgtm

